PR #87, opened 2022-10-19
feat: add dependabot to monitor dependencies

rdimitrov:
The following PR adds Dependabot to monitor for dependency updates

Fixes: https://github.com/vmware/repository-service-tuf/issues/56 and https://github.com/vmware/repository-service-tuf/issues/57

Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+  # Monitor Github Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+    open-pull-requests-limit: 10
+  # Monitor Python test dependencies
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+    open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,16 +7,17 @@ updates:
       interval: "daily"
       time: "10:00"
     commit-message:
-      prefix: "chore"
+      prefix: "build"
       include: "scope"
     open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+updates:
+  # Monitor Github Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
```
kairoaraujo, line 11: IMO is better to use a specific `prefix` for GHA commit messages such as `"gha"`

.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
...
+    open-pull-requests-limit: 10
+  # Monitor Python test dependencies
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
```
kairoaraujo, line 21: Same here, we could have like `prefix`: `requirements`
I also suggest adding the `prefix-development`: `requirements dev`

kairoaraujo (changes requested):
Thank you @rdimitrov !

.github/dependabot.yml
rdimitrov: We can do that, but I thought to not diverge from the general practice - ref. https://www.conventionalcommits.org/en/v1.0.0/#summary 

In that sense, the spec more or less combines everything dependency related under `build`. 

Do we prefer to have everything under `chore`(or `build`) or have separate ones like `gha`, `pip`, `pip dev`, and so forth?

kairoaraujo: Got it!

.github/dependabot.yml
rdimitrov: Done 👍

.github/dependabot.yml
rdimitrov: Settled to `build-dev` for development related dependencies and `build` for everything else 😃

kairoaraujo (approved):

PR #88, opened 2022-10-19
Dev/add UT for ceremony._key_is_duplicated

### LINKED ISSUE #16: Add UT for ``ceremony._key_is_duplicated``
Add UT to the small function `repository_service_tuf.admin.ceremony._key_is_duplicated``

KAUTH:
Adding Unit Tests for ceremony._key_is_duplicated().
- types added to some of the key-related properties of classes
- corrected logic of the _key_is_duplicated function
- fixed typos
- added UTs

Closes #16

### DIFF
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -4,7 +4,7 @@
 
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timedelta
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, TypedDict
 
 from securesystemslib.signer import Signer, SSlibSigner  # type: ignore
 from tuf.api.metadata import (
@@ -33,11 +33,25 @@ repository_metadata: Dict[str, Metadata] = {}
 
 
 @dataclass
+class KeyProp:
+    # "key": Any (Any follows the KEY_SCHEMA from securesystemslib)
+    key: Optional[Dict[str, Any]] = None
+    error: Optional[str] = None
+
+
+@dataclass
+class KeyInput(TypedDict):
+    filename: str
+    password: str
+    key: KeyProp
+
+
+@dataclass
 class RolesKeysInput:
     expiration: int = 1
     num_of_keys: int = 1
     threshold: int = 1
-    keys: Dict[str, Any] = field(default_factory=dict)
+    keys: Dict[str, KeyInput] = field(default_factory=dict)
     offline_keys: bool = True
     paths: Optional[List[str]] = None
     number_hash_prefixes: Optional[int] = None
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -39,7 +39,6 @@ class KeyProp:
     error: Optional[str] = None
 
 
-@dataclass
 class KeyInput(TypedDict):
     filename: str
     password: str
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -2,9 +2,9 @@
 #
 # SPDX-License-Identifier: MIT
 
-from dataclasses import asdict, dataclass, field
+from dataclasses import asdict, dataclass
 from datetime import datetime, timedelta
-from typing import Any, Dict, List, Optional, Tuple, TypedDict
+from typing import Any, Dict, List, Optional, Tuple
 
 from securesystemslib.signer import Signer, SSlibSigner  # type: ignore
 from tuf.api.metadata import (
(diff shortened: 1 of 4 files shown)


rdimitrov:
Hey, thanks for opening the PR, but can you provide a bit more info about what it does?

repository_service_tuf/helpers/tuf.py
```diff
@@ -32,12 +32,25 @@
 repository_metadata: Dict[str, Metadata] = {}
 
 
+@dataclass
+class KeyProp:
+    # "key": Any (Any follows the KEY_SCHEMA from securesystemslib)
+    key: Optional[Dict[str, Any]] = None
+    error: Optional[str] = None
```
kairoaraujo, line 39: I'm not sure if we need this dataclass here.

repository_service_tuf/helpers/tuf.py
```diff
@@ -32,12 +32,25 @@
...
+
+
+class KeyInput(TypedDict):
+    filename: str
+    password: str
+    key: KeyProp
```
kairoaraujo, line 45: This could be a dataclass, and we could use the `asdict` to (de)serialize it. As I mentioned before, it would be more consistent with the already existent dataclasses. Not sure that KeyProp made it more clear.

kairoaraujo (commented):
Thank you @KAUTH

kairoaraujo:
> ceremony._key_is_duplicated). So it's adding Unit Tests for the _key_is_duplicated function. In each commit, there's a detailed commit message regarding the necessary changes.

Optionally, I think you can stash all commits.

rdimitrov:
> > Hey, thanks for opening the PR +1 Can you provide a bit more info about what it does?
> 
> Hey @rdimitrov thanks for responding so quickly! This PR basically addresses #16 (Add UT for ceremony._key_is_duplicated). So it's adding Unit Tests for the _key_is_duplicated function. In each commit there's a detailed commit message regarding the necessary changes.

Thanks, it's awesome that you contribute towards improving the tests 💯 👍 

As a side note, it's good practice to put similar details in the description field of the PR. We'll have PR/Issue templates soon so that will help too 😃

KAUTH:
> > ceremony._key_is_duplicated). So it's adding Unit Tests for the _key_is_duplicated function. In each commit, there's a detailed commit message regarding the necessary changes.
> 
> Optionally, I think you can stash all commits.

Do you mean to squash them? We could do that to keep the commit history a bit more clean-looking. I separated the first commits in order to create some logic segmentation and make the review a bit easier (e.g., first function refactoring and types & then the actual UTs).

repository_service_tuf/helpers/tuf.py
KAUTH: > I'm not sure if we need this dataclass here.
> Not sure that KeyProp made it more clear.

Seeing the code for the first time it was pretty difficult to find out the types of the different class properties and their keys.
e.g., where this ```k.get("key")):``` ([ref](https://github.com/vmware/repository-service-tuf-cli/blob/main/repository_service_tuf/cli/admin/ceremony.py#L295)) or ```k.get("path")``` ([ref](https://github.com/vmware/repository-service-tuf-cli/blob/main/repository_service_tuf/cli/admin/ceremony.py#L297), which by the way the ```path``` key did not exist in the RolesKeysInput.keys.values) keys came from.
I think KeyProp and KeyInput types clear up the types and their structure.

kairoaraujo: Ok, I got your point.
Let's just add an explicit name for `KeyProp`.
I also think making `KeyInput` a dataclass to keep the standard and we can use `asdict` if necessary.
what do you think?

KAUTH: > Let's just add an explicit name for ```KeyProp```.

What do you think about ```KeySchema```?

> I also think making ```KeyInput``` a dataclass to keep the standard and we can use ```asdict``` if necessary.

sounds good! (I didn't know about the asdict method before, nice)

kairoaraujo: `KeySchema` looks good.

IMO, the problem is not that we are replacing typing as an object dataclass.

We make it a dataclass and need to adjust the data structure used in some other places to be consistent.

In `RolesKeyInput`, the `keys` is a required field, and when we initiate the `RolesKeyInput` without giving the keys, we use the default factory as a dict (before, we used a standard python Dictionary, and in your suggestion, we replace by a TypeDict that is compatible)

In practice, we do it in 
https://github.com/vmware/repository-service-tuf-cli/blob/5e9097862f2b98afd5d78196a8d8641213152485/repository_service_tuf/cli/admin/ceremony.py#L287-L290

An option here is make the `keys` field `keys: Optional[Dict[str, KeyInput]] = None`

Neat: I think even the `RolesKeysInput` name is not good anymore; probably, I started as 'key role schema', and it becomes more of a `RoleSettingsInput`.

Also, we need to adjust the usage when we parse the keys.

In the end, our options are
1) We move to dataclass and adjust everything
2) We could keep the dictionary
3) Use TypedDict, but still my last option, as it sounds more an workaround for option 1

KAUTH: Thanks for the elaborate response @kairoaraujo, I see your point on how the classes evolved.
I think it makes sense to try to move to dataclass and adjust everything and it seems like the right time to do it.
I will update you with a new commit push 😉

kairoaraujo (approved):
LGTM

PR #110, opened 2022-10-24
Add pre-commit tool

KAUTH:
In this PR, we add the pre-commit tool to
improve the Software Development Lifecycle
of the project by automatically running checks
(mainly linting & formatting) before one commits their changes.

- Specifically:
* flake8, isort and black were adjusted in tox to run as a pre-commit hooks
* Additionally, added hooks were:
 - the check-added-large-files - prevents giant files from being committed
 - end-of-file-fixer - ensures that a file is either empty or ends with one newline
 - trailing-whitespace - trims trailing whitespace
 - check-yaml - checks yaml files for parseable syntax
* we added a local hook with pre-commit that calls tox to check if `make requirements` is up-to-date

- The Pipfile dev-packages were updated to include pre-commit.

- We updated the README.md with how to install and enable pre-commit.

- We ran `make requirements`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,35 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.3.0
+  hooks:
+  - id: check-added-large-files
+  - id: end-of-file-fixer
+    files: 'repository_service_tuf/'
+  - id: trailing-whitespace
+    files: 'repository_service_tuf/'
+  - id: check-yaml
+    files: '.github/'
+- repo: https://github.com/pycqa/flake8
+  rev: '5.0.4'
+  hooks:
+  - id: flake8
+    exclude: repository_service_tuf/__init__.py|venv|.venv|setting.py|.git|.tox|dist|docs|/*lib/python*|/*egg|build|tools
+- repo: https://github.com/PyCQA/isort
+  rev: '5.10.1'
+  hooks:
+  - id: isort
+    args: [-l79, --profile, black, --check, --diff, .]
+- repo: https://github.com/psf/black
+  rev: '22.10.0'
+  hooks:
+  - id: black
+    args: [-l79, --check, --diff, .]
+- repo: local
+  hooks:
+    - id: tox-requirements
+      name: run requirements check from tox
+      description: Checks if `make requirements` is up-to-date
+      entry: tox -e requirements
+      language: system
+      files: ''
+      verbose: true
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,11 @@ requirements:
 	pipenv requirements > requirements.txt
 	pipenv requirements --dev > requirements-dev.txt
 
+precommit:
+	pre-commit install
+	pre-commit autoupdate
+	pre-commit run --all-files --show-diff-on-failure
+
 coverage:
 	coverage report
 	coverage html -i
--- a/README.rst
+++ b/README.rst
@@ -90,6 +90,21 @@ The flag -d will install the development requirements:
         $ LDFLAGS=-L$(brew --prefix libffi)/lib CFLAGS=-I$(brew --prefix libffi)/include pip install cffi cryptography
 
 
+Running checks with pre-commit:
+
+The pre-commit tool is installed as part of the development requirements.
+
+To automatically run checks before you commit your changes you should run:
+
+.. code:: shell
+
+    $ make precommit
+
+This will install the git hook scripts for the first time, it will update to the
+latest versions of the hooks and run the pre-commit tool.
+Now ``pre-commit`` will run automatically on ``git commit``.
+
+
 Running RSTUF CLI:
 
 .. code:: shell
@@ -130,17 +145,4 @@ Perform automated testing with the tox tool:
 .. code:: shell
 
     $ tox
-
-
-Installing & enabling pre-commit
-================================
-
-The pre-commit tool is installed as part of the development requirements.
-
-To automatically run checks before you commit your changes you should install
-the git hook scripts with **pre-commit**:
-
-.. code:: shell
-
-    $ pre-commit install
 
--- a/README.rst
+++ b/README.rst
@@ -131,3 +131,16 @@ Perform automated testing with the tox tool:
 
     $ tox
 
+
+Installing & enabling pre-commit
+================================
+
+The pre-commit tool is installed as part of the development requirements.
+
+To automatically run checks before you commit your changes you should install
+the git hook scripts with **pre-commit**:
+
+.. code:: shell
+
+    $ pre-commit install
+
--- a/tox.ini
+++ b/tox.ini
@@ -34,8 +34,8 @@ allowlist_externals =
     bash
 commands =
     pipenv --version
-    bash -c 'diff requirements.txt <(pipenv requirements)'
-    bash -c 'diff requirements-dev.txt <(pipenv requirements --dev)'
+    bash -c 'diff -w requirements.txt <(pipenv requirements)'
+    bash -c 'diff -w requirements-dev.txt <(pipenv requirements --dev)'
 
 [testenv:docs]
 deps = -r{toxinidir}/docs/requirements.txt
--- a/tox.ini
+++ b/tox.ini
@@ -5,17 +5,18 @@ skipsdist = true
 
 
 [flake8]
-exclude = ownca/__init__.py,venv,.venv,settings.py,.git,.tox,dist,docs,*lib/python*,*egg,build,tools
+exclude = repository_service_tuf/__init__.py,venv,.venv,settings.py,.git,.tox,dist,docs,*lib/python*,*egg,build,tools
 
 [testenv]
 deps = -r{toxinidir}/requirements.txt
        -r{toxinidir}/requirements-dev.txt
 
 [testenv:lint]
+deps = pre-commit
 commands =
-    flake8
-    isort -l79 --profile black --check --diff .
-    black -l79 --check --diff .
+    pre-commit run flake8 --all-files --show-diff-on-failure
+    pre-commit run isort --all-files --show-diff-on-failure
+    pre-commit run black --all-files --show-diff-on-failure
     # mypy . #issue33
 
 [testenv:test]
--- a/Pipfile
+++ b/Pipfile
@@ -41,6 +41,7 @@ sphinxcontrib-plantuml = "*"
 sphinx-rtd-theme = "*"
 hatchling = "==0.22.0"
 build = "*"
+pre-commit = "*"
 
 [requires]
 python_version = "3.10"
--- a/Pipfile
+++ b/Pipfile
@@ -4,20 +4,13 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-rich-click = "==1.2.1"
-securesystemslib = "==0.22.0"
-cryptography = "==36.0.2"
-cffi = "==1.15.0"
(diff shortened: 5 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+  rev: v4.3.0
+  hooks:
+  - id: check-added-large-files
+  - id: end-of-file-fixer
+    files: 'repository_service_tuf/'
+  - id: trailing-whitespace
+    files: 'repository_service_tuf/'
+  - id: check-yaml
+    files: '.github/'
+- repo: https://github.com/pycqa/flake8
```
kairoaraujo, line 12: Is it required to use the `repo` with version (`rev`) here? Once we already have our Pipenv/requirements-dev?

README.rst
```diff
@@ -131,3 +131,16 @@ Perform automated testing with the tox tool:
 
     $ tox
 
+
+Installing & enabling pre-commit
```
kairoaraujo, line 135: I think it should be part of **Preparing the environment** 

https://github.com/vmware/repository-service-tuf-cli/blob/e0cae3379ef9ee618f3be8ee0934d10a03ec047b/README.rst#preparing-the-environment

README.rst
```diff
@@ -131,3 +131,16 @@ Perform automated testing with the tox tool:
...
+================================
+
+The pre-commit tool is installed as part of the development requirements.
+
+To automatically run checks before you commit your changes you should install
+the git hook scripts with **pre-commit**:
```
kairoaraujo, line 141:
```suggestion
the git hook scripts with ``pre-commit``:
```

kairoaraujo (commented):
Thank you @KAUTH 

I have some questions about the pre-commit tool.
My biggest concern is about how the pre-commit manages the tools.

.pre-commit-config.yaml
KAUTH: Unfortunately, from what I researched it is required (rev is also a required key). requirements-type files and pre-commit-config.yaml don't cooperate together, see https://github.com/pre-commit/pre-commit/issues/730 and https://github.com/pre-commit/pre-commit/issues/945 (accordingly for JS).

There are different mitigation to this:
1) removing these linters from Pipenv/requirements-dev and using [pre-commit autoupdate](https://pre-commit.com/#updating-hooks-automatically) or manual update for the versions (cons: another place to keep track of versions)

2) Running tox from pre-commit as a local hook (see what I did with [tox-requirements hook](https://github.com/vmware/repository-service-tuf-cli/pull/110/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R27)). This way we are installing from tox the versions specified in Pipenv/requirements-dev
cons: we will not be able to take advantage of a key feature of pre-commit; that you can run pre-commit only on the last commit (here I use for the first time pre-commit run --all-files). This is extremely fast and also gives you the chance to slowly migrate changes if you are adapting to another linting format or adding something totally new at some point in the future, without having to refactor the whole code at once for the linter check

kairoaraujo: I think the tool is very nice. I like the idea.

The two points that I'm not fully convinced about are the usage and dependencies on versioning maintenance.
- About the usage: The developer/contributor still need to install (`pre-commit install`) it by himself, having a big chance to them skip it. I say that because many contributors don't run `tox -r`. 
- About the dependency: `.pre-commit-config.yaml`` has its dependency version list, and [it doesn't work with dependabot.](https://github.com/dependabot/dependabot-core/issues/1524)

I would like to hear from @MVrachev and @rdimitrov 😄

rdimitrov (commented):
Can you describe what would change in the steps of creating a commit from a developer's point of view? 

Also how about we wrap the necessary commands in a make target, i.e. something like `make lint` which installs & runs the tools you mention? 

In any case, I like the idea of having a tool that enforces/validates proper formatting so +1 for that 👍

MVrachev:
> About the usage: The developer/contributor still need to install (pre-commit install) it by himself, having a big chance to them skip it. I say that because many contributors don't run tox -r.

We should decide how we want to use `pre-commit` as a local development tool or a tool that we are going to run in CI?
If we are running it locally then if the developer forgets to run `pre-commit install` he can still continue development.

> About the dependency: `.pre-commit-config.yaml`` has its dependency version list, and it doesn't work with dependabot.

Yes, that's true `pre-commit` is not yet supported by dependabot.
One solution to that is to use the [pre-commit autoupdate](https://pre-commit.com/#pre-commit-autoupdate) command to update its dependencies.
We can add this command to the `requirements` command.

KAUTH:
> Can you describe what would change in the steps of creating a commit from a developer's point of view?

Thanks for the response @rdimitrov. 
After you install pre-commit (```pip install pre-commit```, ```pre-commit install```) the developer doesn't have to do anything else. Every time they run ```git commit -s``` and write their commit message the pre-commit hook is run and checks for anything we place there (our linters). If the pre-commit hook fails the commit is not made, enforcing the developer to fix first the lining errors.

> Also how about we wrap the necessary commands in a make target, i.e. something like `make lint` which installs & runs the tools you mention?

That is a good idea, basically, it would be ```pip install pre-commit```, ```pre-commit install```

tox.ini
```diff
@@ -5,17 +5,18 @@ skipsdist = true
...
 
 [testenv:lint]
+deps = pre-commit
 commands =
-    flake8
-    isort -l79 --profile black --check --diff .
-    black -l79 --check --diff .
+    pre-commit run flake8 --all-files --show-diff-on-failure
+    pre-commit run isort --all-files --show-diff-on-failure
+    pre-commit run black --all-files --show-diff-on-failure
```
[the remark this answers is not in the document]
KAUTH: Thanks for the feedback @MVrachev. 

> What's the difference between `pre-commit run flake8 --all-files --show-diff-on-failure` and `flake8` command?
> 
> I am not sure I understand the benefit of having `pre-commit` here instead of directly calling the linters.

This way you have the same exact commands run in the CI as you have locally by the pre-commit hook (run by the developer).

Another positive side-effect is that you can lint, if you choose it to ([it's the default option](https://pre-commit.com/#4-optional-run-against-all-the-files)), only on the changed files.
So if for instance, you are introducing a new tool, such as pylint, and you want to have a gradual migration and not have edits to almost all your project files to conform to the new style, you only need to apply the changes to the files that have been changed on your last commit.

KAUTH:
@kairoaraujo @rdimitrov @MVrachev 
For reference you can take a look at these popular repos and how they use pre-commit alongside tox:
- https://github.com/PyCQA/pylint/blob/main/tox.ini
- https://github.com/asottile/pyupgrade/blob/main/tox.ini
- https://github.com/pallets/flask/blob/main/tox.ini
You will notice that most use it the way suggested in this PR, but there's always the other way proposed, which is to call tox from pre-commit (with the disadvantage I already mentioned in a previous comment, of not being able to run only in the altered files).

kairoaraujo:
@KAUTH, it's LGTM.
Can you please apply the agreed comments?

kairoaraujo (approved):
Doing my last review, I was still a bit confused when I read the README.rst
The unclear part for me is having `tox`, `make precommit` and `make tests` (not explicitly).
I think making a unique and standard path is better, like `make tests` (but giving clear information about what will be happening is good).

I don't have a clear opinion/suggestion on it. I will be approving it but I would like to see another approval from @MVrachev or @rdimitrov before merging 😄

rdimitrov (commented):
> Doing my last review, I was still a bit confused when I read the README.rst The unclear part for me is having `tox`, `make precommit` and `make tests` (not explicitly). I think making a unique and standard path is better, like `make tests` (but giving clear information about what will be happening is good).
> 
> I don't have a clear opinion/suggestion on it. I will be approving it but I would like to see another approval from @MVrachev or @rdimitrov before merging 😄

Yeah, I think we can a bit more info in the README file so we describe these more explicitly. For example, there's already information about how to fork and keep your code in sync, so it would help if we cover the rest of the workflow too, i.e. 
" 
...
_Introduce your changes._
_Verify that code doesn't break and it is formatted properly by running "make tests precommit"_
_Commit_
_Push_
_Open a PR_
...
"
@kairoaraujo @KAUTH - A side note - seeing this, what do you think about having 3 make targets, 1 where these two are called together -i.e. `make tests` and then the rest to be things like `tests-tox` and `tests-precommit`? Would it help make things simpler?

Makefile
```diff
@@ -11,6 +11,11 @@ requirements:
 	pipenv requirements > requirements.txt
 	pipenv requirements --dev > requirements-dev.txt
 
+precommit:
+	pre-commit install
+	pre-commit autoupdate
```
MVrachev, line 16: Don't we want to add the `autoupdate` command in the `requirements` file and check for changes?
The same way we do with `pipenv` and I have suggested in my comment here: https://github.com/vmware/repository-service-tuf-cli/pull/110#issuecomment-1291935416.

MVrachev (commented):
Overall I am fine with these changes.

> Yeah, I think we can have a bit more info in the README file so we describe these more explicitly.

I agree with @KAUTH here that this documentation about `pre-commit` and both sections `Preparing the environment` and `How to add requirements` should be in `CONTRIBUTING.rst`. 
We should focus on information about the project here but I think this should be done in another pr.

> @kairoaraujo @KAUTH - A side note - seeing this, what do you think about having 3 make targets, 1 where these two are called together -i.e. make tests and then the rest to be things like tests-tox and tests-precommit? Would it help make things simpler?

I agree we should rething our MAKEFILE strategy, but probably a separate issue for this will be better as we can easily can get distracted and go into details.

Makefile
KAUTH: Good catch @MVrachev, thanks! I added it in the precommit step of the Makefile but forgot to add it in the requirements as you suggested.
By the way, we will also need to update a bit the description of the update-python-deps.yml workflow, since ```make requirements``` will also include the .pre-commit-config.yaml changes in the PR that it will create. @rdimitrov do you see any problem there, can I open an issue for that?

KAUTH: Good catch @MVrachev, thanks! I added it in the precommit step of the Makefile but forgot to add it in the requirements as you suggested because I was hesitating over the workflow side-effects it could have.
We will also need to update a bit the update-python-deps.yml workflow, since ```make requirements``` will also include the .pre-commit-config.yaml changes in the PR that it will create.
We will need to add pre-commit as an install prerequisite and update the description. @rdimitrov do you see any problem there, can I open an issue for that or resolve it in this PR?

rdimitrov (approved):
@KAUTH - Makes sense 👍 

Alright, let's finalize this one and open additional issues for - 
* Refactor what should be where for README and CONTRIBUTING files
* Rethink the Makefiles targets for testing and linting

Also, keep in mind if you need to update the rest of the PRs with the feedback so far 👍

KAUTH:
> @KAUTH - Makes sense +1
> 
> Alright, let's finalize this one and open additional issues for -
> 
>     * Refactor what should be where for README and CONTRIBUTING files
> 
>     * Rethink the Makefiles targets for testing and linting
> 
> 
> Also, keep in mind if you need to update the rest of the PRs with the feedback so far +1

Good summary @rdimitrov!
I would also add as a note in the Makefile discussion/PR: we might need to update the update-python-deps.yml workflow, if ```make requirements``` will also include the .pre-commit-config.yaml changes in the PR that it will create.
We will need to add pre-commit as an install prerequisite, call pre-commit install (or add that too in make requirements) and update the description.

kairoaraujo (approved):

PR #148, opened 2022-12-12
tests: save metadata when generate the default

kairoaraujo:
This will save the metadata during the UT for bootstrap.

This data can be reused in other tests such as FT

### DIFF
--- a/repository_service_tuf/cli/admin/ceremony.py
+++ b/repository_service_tuf/cli/admin/ceremony.py
@@ -576,8 +576,8 @@ def _bootstrap_state(task_id, server, headers):
     "-s",
     "--save",
     help=(
-        "Save a copy of the metadata localy. This option saves the metadata "
-        "files (json) in the 'metadata' dir."
+        "Save a copy of the metadata locally. This option saves the metadata "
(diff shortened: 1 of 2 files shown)


rdimitrov (approved):
Should we consider deleting it too at some point or that won't be a problem?

MVrachev:
Where will it save this metadata and does the functional tests know exactly where to look for this data?
Do we want to add this data to the list of ignored files by `git` so that they won't be added in the repo, but still can be saved during a call from the functional tests workflow file?

kairoaraujo:
> Should we consider deleting it too at some point or that won't be a problem?

No, it is in the `.gitignore` already. Also the files are overwritten

kairoaraujo:
> Where will it save this metadata and does the functional tests know exactly where to look for this data?
The help of ``--save`` documents it. I fixed and improved its description.

> Do we want to add this data to the list of ignored files by `git` so that they won't be added in the repo, but still can be saved during a call from the functional tests workflow file?

It is already ignored in .gitignore.

MVrachev (approved):
Thanks for the answers @kairoaraujo!

PR #172, opened 2023-01-11
Remove python3.9 and add python3.11 ci build

MVrachev:
I noticed we run build for python3.9 when we say we support only python3.10 onwards.
That's why it seemed unnecessary that we are testing for python3.9.

On the other hand, python 3.11 was introduced back in October 2022 and we can directly add support for it.

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-versions: [ "3.9", "3.10", ]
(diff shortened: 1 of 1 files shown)


rdimitrov (approved):

MVrachev:
I updated the pr as suggested by @kairo in https://github.com/vmware/repository-service-tuf-api/pull/222#pullrequestreview-1244141479

kairoaraujo (approved):

PR #175, opened 2023-01-12
Remove all mentions and usage of the `BIN` role

MVrachev:
As described in the feature https://github.com/vmware/repository-service-tuf/issues/28 we no longer need the `BIN` role.
We can remove all mentions and usages of it and instead use the top-level `targets` role to do its purpose and use succinct hash bin delegation.

Closes https://github.com/vmware/repository-service-tuf-cli/issues/174


tests/unit/cli/admin/test_ceremony.py
```diff
@@ -332,16 +329,13 @@ def test_ceremony_start_default_values(
             "strongPass",
             "tests/files/timestamp1.key",
             "strongPass",
-            "tests/files/JoeCocker.key",
```
MVrachev, line 335: The `tests/files/JoeCocker.key` should be the key we use for the `bin` role.

tests/unit/cli/admin/test_ceremony.py
```diff
@@ -332,16 +329,13 @@ def test_ceremony_start_default_values(
...
-            "strongPass",
             "tests/files/bins1.key",
             "strongPass",
             "y",
             "y",
             "y",
             "y",
             "y",
-            "y",
```
MVrachev, line 344: Removing one `y` as an input can be done as the last six `y` inputs are used to answer on the question wheter a role configuration are correct and given we don't use the `BIN` role anymore we can remove one `y`.

tests/unit/cli/admin/test_ceremony.py
```diff
@@ -403,9 +396,6 @@ def test_ceremony_start_default_values_reconfigure_one_role(
             "",
             "",
             "",
-            "",
-            "",
-            "",
```
MVrachev, line 408: We can remove 3 `""` inputs as to configure one role three questions are asked.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -74,17 +74,17 @@
...
+                                                 .
+                                                 .
+                             .........................................
+                             .                   .                   .
+                             .                   .                   .
+                             .                   .                   .
+                             V                   V                   V
+                      .-------------,     .-------------,     .-------------,
+                      |  bins 0-X   | ... |  bin A-FF   | ... |  bin X-FF   |
+                      `-------------'     `-------------'     `-------------'
```
MVrachev, line 87: I added a third role, so it could be symmetrical.

kairoaraujo (changes requested):

MVrachev:
Updated the examples.

MVrachev:
Changed the target branch to `feature/roles-simplification`.
We plan to combine all role simplifications in this branch as most likely they will be breaking changes.

docs/source/guide/index.rst
```diff
@@ -155,56 +156,50 @@ Step 1: Configure the Roles
...
 
-    • https://example.com/downloads/ is *                                                              
-    • https://example.com/downloads/productA/ is */*                                                   
-    • https://example.com/downloads/productB/ is */* (same as above)                                   
-    • https://example.com/downloads/productB/updates/ is */*/*                                         
+    As we said the targets will be uniformly distributed over the 8 bins no matter if they are
+    located in the same folder. In this example here is how they will be distributed:
+    - "1.bins-0.json" will be responsible for file https://example.com/downloads/productA/productA-v1.0.tar
+    - "1.bins-1.json" will be responsible for file https://example.com/downloads/productB/productB-v1.0.tar
+    - "1.bins-5.json" will be responsible for file https://example.com/downloads/productB/updates/servicepack-1.tar
```
kairoaraujo, line 173: I would remove the base URL and put only the path, to avoid the user understand that the `path` is full url
```suggestion
    - "1.bins-0.json" will be responsible for file productA/productA-v1.0.tar
    - "1.bins-1.json" will be responsible for file productB/productB-v1.0.tar
    - "1.bins-5.json" will be responsible for file productB/updates/servicepack-1.tar
```

kairoaraujo (changes requested):
It looks very good,  only one small change in the guided step.

docs/source/guide/index.rst
MVrachev: Okay, will do and will recalculate where those target roles will be.

MVrachev:
Updated. @kairoaraujo ready for rereview.

kairoaraujo (approved):

PR #181, opened 2023-01-21
Reintroduce mypy checks

### LINKED ISSUE #33: Introduce mypy checks back
The tests are skipping the mypy check, it must to be back before the release.

KAUTH:
Adding a bit of context below since this PR addresses quite a few typing issues.

- In the `_update_snapshot` function we encounter the following typing error: `repository_service_tuf/helpers/tuf.py:195: error: Value of type variable "T" of "Metadata" cannot be "Signed"  [type-var]`. I have ignored it for now, if we can't find a solution in the scope of this PR I suggest opening a new issue to track it, so that we don't block the merging of this PR.

- For errors of the type:
`error: Skipping analyzing <module_name>: module is installed, but missing library stubs or py.typed marker  [import]`, there are basically 4 possible options, see https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker:
1. _Upgrading the version of the library you’re using, in case a newer version has started to include type hints._ - This is not the case for these libraries
2. _Searching to see if there is a PEP 561 compliant stub package. corresponding to your third party library._ - There are none
3. _Writing your own stub files containing type hints for the library._ - This is a substantial extra effort and could be considered in a different PR
4. _Suppress the errors. All this will do is make mypy stop reporting an error on the line containing the import: the imported module will continue to be of type Any._ - Which has the inherited drawback of not actually using types in this case. In this PR we are going with this option

- Excluded type-checking for `docs` directory. Currently, there is only one Python file to be checked there, which is a configuration file, `conf.py`.

- When populating the public key store in `_update_snapshot` function we encounter the error: `repository_service_tuf/helpers/tuf.py: error: Unsupported target for indexed assignment ("Mapping[str, Role]") [index]`.
Here we were assigning a value, `root.roles[role_name] = Role([], threshold)`, but from the `tuf` package `Root.roles` is a `Mapping` which does not support that.
We solve this issue by creating first the `roles` dictionary and initiating the `Root` class with it.

- For `_update_snapshot`, we see the error: `Unsupported target for indexed assignment ("Optional[Dict[str, DelegatedRole]]")  [index]`.
We have to assert that the added dict is not `None`, since the `Delegations.roles` property value is `Optional` in the original `tuf` package.
We encounter similar errors caused by discrepancies in our code and the original package's types (not handling the case that a variable might be `None` since in the original package the variable type includes `Optional`).

Closes #33

### DIFF
--- a/repository_service_tuf/cli/admin/ceremony.py
+++ b/repository_service_tuf/cli/admin/ceremony.py
@@ -426,7 +426,7 @@ def _configure_keys(rolename: str, role: RolesKeysInput) -> None:
         key_count += 1
 
 
-def _check_server(settings):
+def _check_server(settings) -> dict[str, str]:
     server = settings.get("SERVER")
     token = settings.get("TOKEN")
     if server and token:
@@ -457,7 +457,7 @@ def _check_server(settings):
     return headers
 
 
-def _bootstrap(server, headers, json_payload):
+def _bootstrap(server, headers, json_payload) -> Optional[str]:
     task_id = None
     response = request_server(
         server,
@@ -486,7 +486,7 @@ def _bootstrap(server, headers, json_payload):
     return task_id
 
 
-def _bootstrap_state(task_id, server, headers):
+def _bootstrap_state(task_id, server, headers) -> None:
     received_state = []
     while True:
         state_response = request_server(
@@ -583,7 +583,7 @@ def _bootstrap_state(task_id, server, headers):
     is_flag=True,
 )
 @click.pass_context
-def ceremony(context, bootstrap, file, upload, save):
+def ceremony(context, bootstrap, file, upload, save) -> None:
     """
     Start a new Metadata Ceremony.
     """
@@ -695,6 +695,7 @@ def ceremony(context, bootstrap, file, upload, save):
                 )
 
                 if rolename == Roles.TARGETS.value:
+                    assert role.paths is not None
                     delegations_row = (
                         f"\n{SETTINGS.service.targets_base_url}".join(
                             ["", *role.paths]
@@ -737,14 +738,18 @@ def ceremony(context, bootstrap, file, upload, save):
         json_payload: Dict[str, Any] = dict()
 
         json_payload["settings"] = {"service": SETTINGS.service.to_dict()}
-        for role, data in SETTINGS.roles.items():
+        for settings_role, data in SETTINGS.roles.items():
             if data.offline_keys is True:
                 data.keys.clear()
 
             if "roles" not in json_payload["settings"]:
-                json_payload["settings"]["roles"] = {role: data.to_dict()}
+                json_payload["settings"]["roles"] = {
+                    settings_role: data.to_dict()
+                }
             else:
-                json_payload["settings"]["roles"][role] = data.to_dict()
+                json_payload["settings"]["roles"][
+                    settings_role
+                ] = data.to_dict()
 
         json_payload["metadata"] = {
             key: data.to_dict() for key, data in metadata.items()
--- a/repository_service_tuf/cli/admin/ceremony.py
+++ b/repository_service_tuf/cli/admin/ceremony.py
@@ -695,7 +695,7 @@ def ceremony(context, bootstrap, file, upload, save) -> None:
                 )
 
                 if rolename == Roles.TARGETS.value:
-                    assert role.paths is not None
+                    assert role.paths is not None  # nosec B101:assert_used
                     delegations_row = (
                         f"\n{SETTINGS.service.targets_base_url}".join(
                             ["", *role.paths]
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -140,9 +140,7 @@ def initialize_metadata(
         _sign(timestamp, Timestamp.type)
         _add_payload(timestamp, Timestamp.type)
 
-    def _update_snapshot(
-        targets_meta: List[Tuple[str, int]]
-    ) -> Metadata[Snapshot]:
+    def _update_snapshot(targets_meta: List[Tuple[str, int]]) -> int:
         """Loads 'snapshot', updates meta info about passed 'targets'
         metadata, bumps version and expiration, signs and persists. Returns
         new snapshot version, e.g. to update 'timestamp'."""
@@ -162,10 +160,10 @@ def initialize_metadata(
     targets = Targets()
     snapshot = Snapshot()
     timestamp = Timestamp()
-    root = Root()
 
     # Populate public key store, and define trusted signing keys and required
     # signature thresholds for each top-level role in 'root'.
+    roles = {}
     for role_name in TOP_LEVEL_ROLE_NAMES:
         threshold = settings[role_name].threshold
         signers = _signers(role_name)
@@ -181,15 +179,21 @@ def initialize_metadata(
                 f"signing threshold '{threshold}'"
             )
 
-        root.roles[role_name] = Role([], threshold)
-        for signer in signers:
-            root.add_key(
-                Key.from_securesystemslib_key(signer.key_dict), role_name
-            )
+        roles[role_name] = Role([], threshold)
+        keys: list[tuple[Key, str]] = [
+            (Key.from_securesystemslib_key(signer.key_dict), role_name)
+            for signer in signers
+        ]
+
+    root = Root(roles=roles)
+    for key in keys:
+        root.add_key(key[0], key[1])
 
     # Add signature wrapper, bump expiration, and sign and persist
     for role in [targets, snapshot, timestamp, root]:
-        metadata = Metadata(role)
+        # TODO: Create an issue for this mypy error if it's not resolved by the
+        # PR reviews
+        metadata = Metadata(role)  # type: ignore
         _bump_expiry(metadata, role.type)
         _sign(metadata, role.type)
         _add_payload(metadata, role.type)
@@ -201,32 +205,39 @@ def initialize_metadata(
     # Update top-level 'targets' role, to delegate trust for all target files
     # to 'bins' role, defining target path patterns, trusted signing keys and
     # required signature thresholds.
-    targets = _load(Targets.type)
-    targets.signed.delegations = Delegations(keys={}, roles={})
-    targets.signed.delegations.roles[BIN] = DelegatedRole(
-        name=BIN,
-        keyids=[],
-        threshold=settings[BIN].threshold,
-        terminating=False,
-        paths=settings[Targets.type].paths,
-    )
+    targets_metadata = _load(Targets.type)
+    targets_metadata.signed.delegations = Delegations(keys={}, roles={})
+    targets_metadata_roles = {
+        BIN: DelegatedRole(
+            name=BIN,
+            keyids=[],
+            threshold=settings[BIN].threshold,
+            terminating=False,
+            paths=settings[Targets.type].paths,
+        )
+    }
+
+    assert targets_metadata_roles is not None
+
+    targets_metadata.signed.delegations.roles = targets_metadata_roles
 
     for signer in _signers(BIN):
-        targets.signed.add_key(
+        targets_metadata.signed.add_key(
             Key.from_securesystemslib_key(signer.key_dict), BIN
         )
 
     # Bump version and expiration, and sign and persist updated 'targets'.
-    _bump_version(targets)
-    _bump_expiry(targets, Targets.type)
-    _sign(targets, Targets.type)
-    _add_payload(targets, Targets.type)
+    _bump_version(targets_metadata)
+    _bump_expiry(targets_metadata, Targets.type)
+    _sign(targets_metadata, Targets.type)
+    _add_payload(targets_metadata, Targets.type)
 
-    targets_meta.append((Targets.type, targets.signed.version))
+    targets_meta.append((Targets.type, targets_metadata.signed.version))
 
-    succinct_roles = SuccinctRoles(
-        [], 1, settings[BINS].number_hash_prefixes, BINS
-    )
+    bit_length = settings[BINS].number_hash_prefixes
+    assert bit_length is not None
+
+    succinct_roles = SuccinctRoles([], 1, bit_length, BINS)
     # Create new 'bins' role and delegate trust from 'bins' for all target
     # files to 'bin-n' roles based on file path hash prefixes, a.k.a hash bin
     # delegation.
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -156,11 +156,6 @@ def initialize_metadata(
 
         return snapshot.signed.version
 
-    # Bootstrap default top-level metadata to be updated below if necessary
-    targets = Targets()
-    snapshot = Snapshot()
-    timestamp = Timestamp()
-
     # Populate public key store, and define trusted signing keys and required
     # signature thresholds for each top-level role in 'root'.
     roles: dict[str, Role] = {}
@@ -185,18 +180,23 @@ def initialize_metadata(
             for signer in signers
         ]
 
-    root = Root(roles=roles)
-    for arg in add_key_args:
-        root.add_key(arg[0], arg[1])
-
     # Add signature wrapper, bump expiration, and sign and persist
-    for role in [targets, snapshot, timestamp, root]:
-        # TODO: Create an issue for this mypy error if it's not resolved by the
-        # PR reviews
-        metadata = Metadata(role)  # type: ignore
-        _bump_expiry(metadata, role.type)
-        _sign(metadata, role.type)
-        _add_payload(metadata, role.type)
+    for role in [Targets, Snapshot, Timestamp, Root]:
+
+        # Bootstrap default top-level metadata to be updated below if necessary
+        if role is Root:
+            metadata = Metadata(role(roles=roles))
+            root = metadata.signed
+            for arg in add_key_args:
+                root.add_key(arg[0], arg[1])
+
+        else:
+            metadata = Metadata(role())
+
+        metadata_type = metadata.signed.type
+        _bump_expiry(metadata, metadata_type)
+        _sign(metadata, metadata_type)
+        _add_payload(metadata, metadata_type)
 
     # Track names and versions of new and updated targets for 'snapshot'
     # update
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -217,7 +217,7 @@ def initialize_metadata(
         )
     }
 
-    assert targets_metadata_roles is not None
+    assert targets_metadata_roles is not None  # nosec B101:assert_used
 
     targets_metadata.signed.delegations.roles = targets_metadata_roles
 
@@ -235,7 +235,7 @@ def initialize_metadata(
     targets_meta.append((Targets.type, targets_metadata.signed.version))
 
     bit_length = settings[BINS].number_hash_prefixes
-    assert bit_length is not None
+    assert bit_length is not None  # nosec B101:assert_used
 
     succinct_roles = SuccinctRoles([], 1, bit_length, BINS)
     # Create new 'bins' role and delegate trust from 'bins' for all target
--- a/repository_service_tuf/helpers/tuf.py
+++ b/repository_service_tuf/helpers/tuf.py
@@ -185,7 +185,7 @@ def initialize_metadata(
 
         # Bootstrap default top-level metadata to be updated below if necessary
         if role is Root:
-            metadata = Metadata(role(roles=roles))
+            metadata = Metadata(Root(roles=roles))
             root = metadata.signed
             for arg in add_key_args:
                 root.add_key(arg[0], arg[1])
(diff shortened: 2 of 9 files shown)


repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -426,7 +426,7 @@ def _configure_keys(rolename: str, role: RolesKeysInput) -> None:
         key_count += 1
 
 
-def _check_server(settings):
+def _check_server(settings) -> dict[str, str]:
```
MVrachev, line 429: Looking into this function I realized that we don't need to have the `else` case in line 445 as if it enters in one of the previous `if` statements the function execution will stop.
If you want you can simplify the function.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -457,7 +457,7 @@ def _check_server(settings):
     return headers
 
 
-def _bootstrap(server, headers, json_payload):
+def _bootstrap(server, headers, json_payload) -> Optional[str]:
```
MVrachev, line 460: I think we can do the same simplification as in `_check_server` and remove the `else` statement on line 481.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -737,14 +738,18 @@ def ceremony(context, bootstrap, file, upload, save):
         json_payload: Dict[str, Any] = dict()
 
         json_payload["settings"] = {"service": SETTINGS.service.to_dict()}
-        for role, data in SETTINGS.roles.items():
+        for settings_role, data in SETTINGS.roles.items():
```
MVrachev, line 741: One of the places where we see why `mypy` checks are really useful.

repository_service_tuf/helpers/tuf.py
```diff
@@ -140,9 +140,7 @@ def _update_timestamp(snapshot_version: int):
         _sign(timestamp, Timestamp.type)
         _add_payload(timestamp, Timestamp.type)
 
-    def _update_snapshot(
-        targets_meta: List[Tuple[str, int]]
-    ) -> Metadata[Snapshot]:
+    def _update_snapshot(targets_meta: List[Tuple[str, int]]) -> int:
```
MVrachev, line 143: I am honestly not sure why here we used `Metadata[Snapshot]` as a return type when we do return an int - the new snapshot version.

repository_service_tuf/helpers/tuf.py
```diff
@@ -162,10 +160,10 @@ def _update_snapshot(
     targets = Targets()
     snapshot = Snapshot()
     timestamp = Timestamp()
-    root = Root()
 
     # Populate public key store, and define trusted signing keys and required
     # signature thresholds for each top-level role in 'root'.
+    roles = {}
```
MVrachev, line 166: If we are going to use type annotations then it's a good practice for dictionaries to write their corresponding internal types (as you did for the `keys` list on line 182) or:
```suggestion
    roles: dict[str, Metadata[T]]` = {}
```
or if `Metadata[T]` is not accepted then just use `Metadata`.

repository_service_tuf/helpers/tuf.py
```diff
@@ -180,15 +178,21 @@ def _update_snapshot(
...
-        root.roles[role_name] = Role([], threshold)
-        for signer in signers:
-            root.add_key(
-                Key.from_securesystemslib_key(signer.key_dict), role_name
-            )
+        roles[role_name] = Role([], threshold)
+        keys: list[tuple[Key, str]] = [
+            (Key.from_securesystemslib_key(signer.key_dict), role_name)
+            for signer in signers
+        ]
```
MVrachev, line 185: Are you sure the elements of `keys` are `tuples`?
As here on lines 183-184 I see that we create only a `Key` instance.

repository_service_tuf/helpers/tuf.py
```diff
@@ -180,15 +178,21 @@ def _update_snapshot(
...
+    root = Root(roles=roles)
+    for key in keys:
+        root.add_key(key[0], key[1])
 
     # Add signature wrapper, bump expiration, and sign and persist
     for role in [targets, snapshot, timestamp, root]:
-        metadata = Metadata(role)
+        # TODO: Create an issue for this mypy error if it's not resolved by the
+        # PR reviews
+        metadata = Metadata(role)  # type: ignore
```
MVrachev, line 195: Really not sure why you have `mypy` issue for this.
Strange.
@lukpueh any idea why we have that issue?
Asking you as this seems to be connected to `python-tuf` `Generic[T]` type.

repository_service_tuf/helpers/tuf.py
```diff
@@ -200,32 +204,39 @@ def _update_snapshot(
...
-    targets.signed.delegations.roles[BIN] = DelegatedRole(
-        name=BIN,
-        keyids=[],
-        threshold=settings[BIN].threshold,
-        terminating=False,
-        paths=settings[Targets.type].paths,
-    )
+    targets_metadata = _load(Targets.type)
+    targets_metadata.signed.delegations = Delegations(keys={}, roles={})
+    targets_metadata_roles = {
```
MVrachev, line 209: Again will prefer it if you point out the dictionary's internal type.

MVrachev (changes requested):
> In the _update_snapshot function we encounter the following typing error: repository_service_tuf/helpers/tuf.py:195: error: Value of type variable "T" of "Metadata" cannot be "Signed"  [type-var]. I have ignored it for now, if we can't find a solution in the scope of this PR I suggest opening a new issue to track it, so that we don't block the merging of this PR.

I am not sure this is an issue as I see a mistake in our current type annotation

> - For errors of the type:
`error: Skipping analyzing <module_name>: module is installed, but missing library stubs or py.typed marker  [import]`, there are basically 4 possible options, see https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker:

I think we can solve this problem by overriding the problematic modules as you did in `[[tool.mypy.overrides]]` right?

repository_service_tuf/helpers/tuf.py
KAUTH: Yeah, I think it was just a stray mistake, perhaps influenced by the fact that the `_load` method returns a `Metadata` type.

repository_service_tuf/helpers/tuf.py
KAUTH: Good point!

repository_service_tuf/cli/admin/ceremony.py
KAUTH: If it's just a nit, I would leave it like this. I have noticed it as a coding style being done elsewhere as well and it may be considered by some more explicit, easier-to-read, etc.

repository_service_tuf/helpers/tuf.py
KAUTH: Hm I can see how this can be confusing. If you see `root.add_key` it takes as arguments a `key: Key` and a `role: str`. So, it's one key, with the mapping "key: role" (what I wrote as key[0] which is the key, and key[1] which is the role). I could have used a dict to make it clearer but the Key class would need to be a frozen class in order to use it as a key.

I will change the naming to make it clearer.

repository_service_tuf/helpers/tuf.py
lukpueh: Looks like we need to annotate the list used in the for loop `[targets, snapshot, timestamp, root]`...

lukpueh: Hm. I know what the problem is, but not how to fix it. :) Btw. error message is:
```
repository_service_tuf/helpers/tuf.py:195: error: Value of type variable "T" of "Metadata" cannot be "Signed"  [type-var]
```

So it looks like the typing system sees a list of different objects with a common base class and probably annotates it as something like `List[Signed]`. But Metadata requires its first argument to be not just any `Signed`, but one of the currently available `Signed` subclasses aka `T`.

My naive approach would be to declare `role: T` or  `roles: List[T]`, but then mypy complains about `T` being unbound.

🤷

jku: > it looks like the typing system sees a list of different objects with a common base class and probably annotates it as something like List[Signed]

yes, you can confirm this by adding `reveal_type(role)` inside the loop (that will only work during mypy)

I don't think there's a clean way to handle this but some ideas:
* could annotate as `role: Union[Root, ...]` but that likely just makes mypy not do anything with it -- the mypy ignore is likely just as useful
* instead of having e.g. a `root=Root()` as the modified object initialize `root_md: Metadata(Root())` instead in the beginning (and then access root_md.signed if you need to -- it's annotated so works just fine)

I would go for the second option. Yes, it's annoying to have to change the flow you had in mind for annotations but I think that style usually works just as well and you get working annotations.

jku: oh that's interesting. I suppose the for loop mis-typing still breaks it...

> Do you know why we can't just do role: T?

T is not a "real type" IIRC: it doesn't exist at runtime in anyway and so can only be used in a function signature (and then _maybe_ within that function, I don't recall the details).

In any case, I suggest trying  `Metadata(Root())`.

MVrachev: > In any case, I suggest trying Metadata(Root()).

Agree, that looks like the best solution.

Wanted to say it's really nice to see you both in a discussion here.
Felt like a discussion on a `python-tuf` contribution for a moment. 😁

repository_service_tuf/helpers/tuf.py
MVrachev: Thanks, now it's clearer after your explanation and the renaming.

repository_service_tuf/helpers/tuf.py
```diff
@@ -200,32 +204,39 @@ def _update_snapshot(
...
+            keyids=[],
+            threshold=settings[BIN].threshold,
+            terminating=False,
+            paths=settings[Targets.type].paths,
+        )
+    }
+
+    assert targets_metadata_roles is not None
+
+    targets_metadata.signed.delegations.roles = targets_metadata_roles
```
MVrachev, line 221: You will have to update this part when you rebase, sorry about that.
We have removed the `bins` role as a whole and now `targets` is using `succinct hash bin delegation`:
https://github.com/vmware/repository-service-tuf-cli/pull/175

tox.ini
```diff
@@ -17,7 +17,10 @@ commands =
     pre-commit run flake8 --all-files --show-diff-on-failure
     pre-commit run isort --all-files --show-diff-on-failure
     pre-commit run black --all-files --show-diff-on-failure
-    # mypy . #issue33
+
+[testenv:typing]
+commands =
+    mypy .
```
MVrachev, line 23: Why not run `mypy` on each pr if we are going to introduce linting?
@kairoaraujo do you agree?
If @KAUTH has spent so much time on this (and I personally strongly support this change) why not move to `mypy` linting for each of our changes as well?

MVrachev (changes requested):

tox.ini
kairoaraujo: Yes, I agree. 🎉 

add `typing` to 
https://github.com/vmware/repository-service-tuf-cli/blob/3e500ea55b00272d197b1330fd75ca3262eefc22/tox.ini#L3
and
https://github.com/vmware/repository-service-tuf-cli/blob/3e500ea55b00272d197b1330fd75ca3262eefc22/tox.ini#L48-L51

@KAUTH integrate also to pre-commit 🚀

repository_service_tuf/helpers/tuf.py
KAUTH: Thanks for the very useful input everyone, ` Metadata(Root())` seems to be doing the trick!

tox.ini
KAUTH: Since the official `mypy` repo doesn't provide a `pre-commit` hook we have to resort to [`mirrors-mypy`](https://github.com/pre-commit/mirrors-mypy) which runs `mypy` from an isolated virtualenv and is not as flexible as `mypy` (e.g., cannot exclude specific modules as we do in the `pyproject.toml` mypy config).
For this reason I am creating a local hook to run `mypy` from.

kairoaraujo:
@KAUTH, #191 is also fixing your previous commit.

KAUTH:
> @KAUTH, #191 is also fixing your previous commit.

Oh, I just noticed it. I approved the PR, I can rebase on top of that change

repository_service_tuf/helpers/tuf.py
```diff
@@ -185,18 +180,23 @@ def _update_snapshot(targets_meta: List[Tuple[str, int]]) -> int:
...
-        # PR reviews
-        metadata = Metadata(role)  # type: ignore
-        _bump_expiry(metadata, role.type)
-        _sign(metadata, role.type)
-        _add_payload(metadata, role.type)
+    for role in [Targets, Snapshot, Timestamp, Root]:
+
+        # Bootstrap default top-level metadata to be updated below if necessary
+        if role is Root:
+            metadata = Metadata(role(roles=roles))
```
MVrachev, line 188: I think you can replace `role` with `Root` to make it clearer:
```suggestion
            metadata = Metadata(Root(roles=roles))
```

MVrachev (commented):
Left a couple of small comments.

repository_service_tuf/helpers/tuf.py
```diff
@@ -217,7 +217,7 @@ def _update_snapshot(targets_meta: List[Tuple[str, int]]) -> int:
         )
     }
 
-    assert targets_metadata_roles is not None
+    assert targets_metadata_roles is not None  # nosec B101:assert_used
```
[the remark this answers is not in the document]
KAUTH: Right, I think I was getting some error related that `targets_metadata.signed.delegations.roles` was Optional, I am not sure, but indeed we can remove it.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -695,7 +695,7 @@ def ceremony(context, bootstrap, file, upload, save) -> None:
                 )
 
                 if rolename == Roles.TARGETS.value:
-                    assert role.paths is not None
```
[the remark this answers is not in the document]
KAUTH: If we don't check somehow whether `role.paths` is not `None`, `mypy` returns the following error: `repository_service_tuf/cli/admin/ceremony.py:701: error: List or tuple expected as variadic arguments  [misc]`

From `_configure_role(rolename, role)` we can see that `role.paths` will not be `None` (`role.paths = delegate_paths`), but `mypy` does not pick that up, thus with this assert we let the static typechecker know that it's not `None`. Therefore, we do not mind whether the `assert` is removed when the code is compiled (it's only used for `mypy`).

This could be checked with an `if` but you would need to handle the case and as I mentioned above I don't see this actually being a possible scenario.

repository_service_tuf/helpers/tuf.py
```diff
@@ -235,7 +235,7 @@ def _update_snapshot(targets_meta: List[Tuple[str, int]]) -> int:
     targets_meta.append((Targets.type, targets_metadata.signed.version))
 
     bit_length = settings[BINS].number_hash_prefixes
-    assert bit_length is not None
```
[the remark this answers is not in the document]
KAUTH: From what I saw (https://github.com/vmware/repository-service-tuf-cli/blob/main/repository_service_tuf/cli/admin/ceremony.py#L385) the default is 8, so it's not possible, at least for the `bins` role.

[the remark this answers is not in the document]
MVrachev: Yes, that's my point. if `bit_length` is `None` for `BINS` or if:
`settings[BINS].number_hash_prefixes` is `None` we should probably throw an exception. There is no valid scenario when `number_hash_prefixes` will be `None` for `BINS`.

repository_service_tuf/cli/admin/ceremony.py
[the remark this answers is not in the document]
MVrachev: Okay, here it seems that this is a valid case for an `assert`.
We know that `role.paths` will not be `None` from `_configure_role(rolename,role)`, but we want to add this assert because of `mypy`.

That's my point in raising the question: do we need an assert here or we want to use an actual `if` statement? Asserts are great if you know that certain cases cannot happen, but you want `mypy` to be happy, otherwise, it's better to use actual `if` statements.

MVrachev (commented):
I have only one single point I want us to discuss:
https://github.com/vmware/repository-service-tuf-cli/pull/181#discussion_r1093093273
besides that point, it looks good to me.

repository_service_tuf/helpers/tuf.py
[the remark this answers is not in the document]
KAUTH: > . if bit_length is None for BINS or if:
settings[BINS].number_hash_prefixes is None we should probably throw an exception

but it's never `None`, as per the link I shared in the code there's always a default int value and `click` checks that any user-input is an int (`IntPrompt`). IMO we should not check whether it's `None` as it is not necessary (the value will never be `None` with the current code) and I would argue that if we did do this for this input we would have to do this for every input given by the user.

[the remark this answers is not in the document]
MVrachev: You convinced me.
I wasn't sure that `click` will check that the user input is an int. This combined with python-tuf SuccinctRoles check for `bit_length` that the number is between 0 and 32: 
https://github.com/theupdateframework/python-tuf/blob/d2c12f2d144b25313cfde68a4c84c2b44b73bd72/tuf/api/metadata.py#L1333
is enough.
We should be wary of user input and I think we need to do a similar analysis for each of our arguments.
We know that the CLI users would not be attackers as they won't have access to the keys.
If they do have access to the keys the problem is already there no matter what we do with the CLI.
What we want to protect our users from is a mistake and given it's a user input from the terminal I think that's important.
I will create an issue about this.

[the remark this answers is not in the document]
MVrachev: https://github.com/vmware/repository-service-tuf-cli/issues/193

MVrachev (approved):
My comments were addressed and I am happy with the changes.

@kairoaraujo do you want to do another review or add anything or I can merge?
Asking as you were involved in the discussions.

repository_service_tuf/helpers/tuf.py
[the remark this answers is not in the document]
KAUTH: > What we want to protect our users from is a mistake and given it's a user input from the terminal I think that's important.

That is a very good point, thanks for also opening an issue

kairoaraujo:
> My comments were addressed and I am happy with the changes.
> 
> @kairoaraujo do you want to do another review or add anything or I can merge? Asking as you were involved in the discussions.

I don't have any comments. LGTM :)

repository_service_tuf/helpers/tuf.py
```diff
@@ -181,18 +174,29 @@ def _update_snapshot(
                 f"signing threshold '{threshold}'"
             )
 
-        root.roles[role_name] = Role([], threshold)
-        for signer in signers:
-            root.add_key(
-                Key.from_securesystemslib_key(signer.key_dict), role_name
-            )
+        roles[role_name] = Role([], threshold)
+        add_key_args: list[tuple[Key, str]] = [
```
KAUTH, line 178: Can we postpone merging, I am not very confident about this line. This is in a for loop for every role name in `TOP_LEVEL_ROLE_NAMES`, but we are re-writing to the `add_key_args` variable before passing them to `root.add_key`

kairoaraujo: Functional tests are good.
The important part is the signing and bootstrap, which comes from the CLI. They are OK. The overall tests are also passing.
```shell
❯ pytest -vv ../repository-service-tuf-cli/tests/unit/cli/admin/test_ceremony.py::TestCeremonyGroupCLI::test_ceremony_start_default_values
========================================================================== test session starts ===========================================================================
platform darwin -- Python 3.10.0, pytest-7.2.1, pluggy-1.0.0 -- /Users/kdearaujo/.local/share/virtualenvs/repository-service-tuf-E73fsQ7i/bin/python
cachedir: .pytest_cache
metadata: {'Python': '3.10.0', 'Platform': 'macOS-12.6.1-x86_64-i386-64bit', 'Packages': {'pytest': '7.2.1', 'pluggy': '1.0.0'}, 'Plugins': {'html': '3.2.0', 'bdd-html': '0.1.14a0', 'metadata': '2.0.4', 'bdd': '6.1.1'}}
rootdir: /Users/kdearaujo
plugins: html-3.2.0, bdd-html-0.1.14a0, metadata-2.0.4, bdd-6.1.1
collected 1 item                                                                                                                                                         

../repository-service-tuf-cli/tests/unit/cli/admin/test_ceremony.py::TestCeremonyGroupCLI::test_ceremony_start_default_values PASSED                               [100%]

=========================================================================== 1 passed in 0.79s ============================================================================
❯ time rstuf -c rstuf.ini admin ceremony -b -u -s -f payload.json
Starting online bootstrap
Bootstrap status: ACCEPTED (847f723b10944f19bde3457cc82c5785)
Bootstrap status: STARTED
Bootstrap status: SUCCESS
Bootstrap finished.

Ceremony done. 🔐 🎉
rstuf -c rstuf.ini admin ceremony -b -u -s -f payload.json  0.38s user 0.14s system 12% cpu 4.235 total
❯ make functional-tests
===================================================================== 68 passed in 368.66s (0:06:08) =====================================================================
```

KAUTH: Alright then, thanks ☺️

kairoaraujo (approved):

MVrachev (approved):

PR #191, opened 2023-01-30
fix `make requirement` and update dev dependencies

### LINKED ISSUE #184: Other: `make requirements` does not update `Pipfile.lock` when `Pipfile` has been manually edited
### What do you want to share with us?

Running `make requirements` does not update the requirement files (`requirements.txt` and `requirements-dev.txt`) correctly if we have manually edited the `Pipfile`. This could happen if we are manually pinning a dependency or specifying some optional dependency.
Locking first could protect us from having outdated requirement files if we manually edit the `Pipfile` and would be a nice-to-have extra precaution.

### Steps to reproduce
- Update the `Pipfile` by adding a new package
- Run `make requirements` to update the requirement files

### References

We expect the requirements files to include the latest changes made to the Pipfile.
This does not happen since we need to lock before running:
```shell
	pipenv requirements > requirements.txt
	pipenv requirements --dev > requirements-dev.txt
```
so we could instead have:
```shell
	pipenv lock
	pipenv requirements > requirements.txt
	pipenv requirements --dev > requirements-dev.txt
```

This is because the `Pipfile.lock` is used to generate the requirement files
> Given the Pipfile.lock exists, you may generate a set of requirements out of it with the default dependencies:

source: https://pipenv.pypa.io/en/latest/advanced/#generating-a-requirements-txt

### Code of Conduct

kairoaraujo:
- fix `make requirements`
- remove black as pinned
- update dev dependencies
- update `.pre-commit-config.yaml` aligned to `requirements-dev.txt`

Closes #184 

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,17 +10,17 @@ repos:
   - id: check-yaml
     files: '.github/'
 - repo: https://github.com/pycqa/flake8
-  rev: '5.0.4'
+  rev: '6.0.0'
   hooks:
   - id: flake8
     exclude: repository_service_tuf/__init__.py|venv|.venv|setting.py|.git|.tox|dist|docs|/*lib/python*|/*egg|build|tools
 - repo: https://github.com/PyCQA/isort
-  rev: '5.10.1'
+  rev: '5.12.0'
   hooks:
   - id: isort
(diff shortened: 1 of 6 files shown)


KAUTH (approved):
LGTM!

Issue #39, opened 2022-08-19
Task: CLI artifact add command `rstuf artifact add`


kairoaraujo:
Implement to the rstuf CLI the functionality to add artifacts.

Two specific features to add targets:

## Add Artifact (Feature)

This target adds the target using the RSTUF REST API. This feature is an interface to add targets using the command line.

### 1. Simple one artifact file 

Note: the user can use the token (`-t/--token`) when API requires authentication/authorization, if the user uses `rstuf --auth` it ignores the token parameter. 

#### Giving the artifact
`rstuf artifact add <file> [-p/--path <file-path>, -c/--custom, -t/--token <token>]`

Example: `rstuf artifact add file-v1.0.3.tar.gz -p projectA/download/`

It gets from the file the hash, size and add a custom path in the metadata `projectA/download/file-v1.0.3.tar.gz`

#### Without file, just the information

`rstuf artifact add -p/--path [file-path], -l/lenght [length] -h/hashes [b2sum] -c/--custom`

### 2. From JSON
`rstuf artifact add -d @file.json`

```json
{
  "targets": [
    {
      "info": {
        "length": "int",
        "hashes": {
          "blake2b-256": "str"
        },
        "custom": {
          "key": "value"
        }
      },
      "path": "str"
    },
    {
      "info": {
        "length": "int",
        "hashes": {
          "blake2b-256": "str"
        },
        "custom": {
          "key": "value"
        }
      },
      "path": "str"
    },
  ]
}
```

It should return the task id.

### Import targets (Feature) 

- [DONE] #195 

This feature adds a huge amount of targets, importing from CSV directly to RSTUF Database
The feature description is detailed in these links
- https://github.com/vmware/repository-service-tuf/issues/188
- BDD Feature: https://github.com/vmware/repository-service-tuf/pull/218

kairoaraujo:
- Add a feature `rstuf admin add-target` option to import directly to DB. It is related to https://github.com/vmware/repository-service-tuf/issues/188

> CLI Implementation
>
> The RSTUF user will be required to create a CSV file that contains the following information
path;size;hash algorithm;hash;
Example:
test/optimistic_lamport-v1.0.tar.gz;12345;blake2b-256;716f6e863f744b9ac22c97ec7b76ea5f5908bc5b2f67c61510bfc4751384ea7a;
>
> The CSV file limit will be half-million of targets (500,000 lines)
>
[...]

PR #195, opened 2023-02-08
feature: import targets

kairoaraujo:
Implements the feature import target `rstuf admin import-targets`.

This feature gives the RSTUF administrator the functionality to import a large number of existent targets. It helps to roll out and deploy the RSTUF in existing repositories.

This feature is detailed and explained at these links
- https://github.com/vmware/repository-service-tuf/issues/188
- BDD Feature: https://github.com/vmware/repository-service-tuf/pull/218

Some changes in the ceremony was done to reuse in the code:

  - The `ceremony._check_server` was converted to `helpers.api_client.get_headers`
  - The `ceremony._bootstrap_state` was converted to `helpers.api_client.task_status`

Added 100% coverage to `helpsers.api_client`

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,6 +13,8 @@ requests = "*"
 tuf = "==2.0.0"
 dynaconf = {extras = ["ini"], version = "*"}
 isort = "*"
+sqlalchemy = "*"
+psycopg2 = "*"
 
 [dev-packages]
 black = "*"
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -14,10 +14,16 @@ from repository_service_tuf.helpers.api_client import (
     Methods,
     get_headers,
     is_logged,
+    publish_targets,
     request_server,
     task_status,
 )
-from repository_service_tuf.helpers.tuf import Metadata, SuccinctRoles
+from repository_service_tuf.helpers.tuf import (
+    Delegations,
+    Metadata,
+    SuccinctRoles,
+    Targets,
+)
 
 console = Console()
 
@@ -80,6 +86,25 @@ def _import_csv_to_rstuf(
         console.print(f"Import status: {csv_file} imported")
 
 
+def _get_succinct_roles(metadata_url: str) -> SuccinctRoles:
+    response = request_server(metadata_url, "1.bin.json", Methods.get)
+    if response.status_code == 404:
+        raise click.ClickException("RSTUF Metadata Targets not found.")
+
+    json_data = json.loads(response.text)
+    targets: Metadata[Targets] = Metadata.from_dict(json_data)
+    if targets.signed.delegations is None:
+        raise click.ClickException("Failed to get Targets Delegations")
+
+    targets_delegations: Delegations = targets.signed.delegations
+
+    if targets_delegations.succinct_roles is None:
+        raise click.ClickException("Failed to get Targets succinct roles")
+
+    succinct_roles: SuccinctRoles = targets_delegations.succinct_roles
+    return succinct_roles
+
+
 @admin.command()
 @click.option(
     "-metadata-url",
@@ -107,7 +132,7 @@ def _import_csv_to_rstuf(
 )
 @click.pass_context
 def import_targets(
-    context,
+    context: Any,
     metadata_url: str,
     db_uri: str,
     csv: List[str],
@@ -131,14 +156,22 @@ def import_targets(
 
     headers = get_headers(settings)
 
-    response = request_server(metadata_url, "1.bin.json", Methods.get)
-    if response.status_code == 404:
-        raise click.ClickException("RSTUF Metadata Targets not found.")
+    bs_response = request_server(
+        settings.SERVER, URL.bootstrap.value, Methods.get, headers=headers
+    )
+    if bs_response.status_code != 200:
+        raise click.ClickException(
+            f"Error {bs_response.status_code} {bs_response.text}"
+        )
+
+    bs_data = bs_response.json()["data"]
+    if bs_data.get("bootstrap") is False:
+        raise click.ClickException(
+            "`import-targets` requires bootstrap process done."
+        )
 
     # load all required infrastructure
-    json_data = json.loads(response.text)
-    targets = Metadata.from_dict(json_data)
-    succinct_roles = targets.signed.delegations.succinct_roles
+    succinct_roles = _get_succinct_roles(metadata_url)
     engine = create_engine(f"{db_uri}")
     db_metadata = MetaData()
     db_client = engine.connect()
@@ -156,21 +189,13 @@ def import_targets(
 
     if skip_publish_targets:
         console.print(
-            "Import status: Finshed. "
+            "Import status: Finished. "
             "Not targets published (`--skip-publish-targets`)"
         )
     else:
         console.print("Import status: Submitting action publish targets")
-        publish_targets = request_server(
-            server, URL.publish_targets.value, Methods.post, headers=headers
-        )
-        if publish_targets.status_code != 202:
-            raise click.ClickException(
-                f"Failed to publish targets. {publish_targets.text}"
-            )
-        task_id = publish_targets.json()["data"]["task_id"]
+        task_id = publish_targets(server, headers)
         console.print(f"Import status: Publish targets task id is {task_id}")
-
         # monitor task status
         result = task_status(task_id, server, headers, "Import status: task ")
         if result is not None:
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -106,7 +106,13 @@ def _import_csv_to_rstuf(
     help="Skip publishing targets in TUF Metadata.",
 )
 @click.pass_context
-def import_targets(context, metadata_url, db_uri, csv, skip_publish_targets):
+def import_targets(
+    context,
+    metadata_url: str,
+    db_uri: str,
+    csv: List[str],
+    skip_publish_targets: bool,
+):
     """
     Import targets to RSTUF from exported CSV file.
     """
@@ -139,7 +145,7 @@ def import_targets(context, metadata_url, db_uri, csv, skip_publish_targets):
     rstuf_table = Table("rstuf_targets", db_metadata, autoload_with=engine)
 
     # validate if the CSV files are accessible
-    _check_csv_files(csv)
+    _check_csv_files(csv_files=csv)
     # import all CSV file(s) data to RSTUF DB without commiting
     _import_csv_to_rstuf(db_client, rstuf_table, csv, succinct_roles)
 
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -100,7 +100,7 @@ def _import_csv_to_rstuf(
 @click.option(
     "--skip-publish-targets",
     is_flag=True,
-    help="Skip publish targets process in TUF Metadata.",
+    help="Skip publishing targets in TUF Metadata.",
 )
 @click.pass_context
 def import_targets(context, metadata_url, db_uri, csv, skip_publish_targets):
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -13,7 +13,6 @@ from repository_service_tuf.helpers.api_client import (
     URL,
     Methods,
     get_headers,
-    is_logged,
     publish_targets,
     request_server,
     task_status,
@@ -106,17 +105,17 @@ def _get_succinct_roles(metadata_url: str) -> SuccinctRoles:
 
 @admin.command()
 @click.option(
-    "-metadata-url",
+    "--metadata-url",
     required=True,
     help="RSTUF Metadata URL i.e.: http://127.0.0.1 .",
 )
 @click.option(
-    "-db-uri",
+    "--db-uri",
     required=True,
     help="RSTUF DB URI. i.e.: postgresql://postgres:secret@127.0.0.1:5433",
 )
 @click.option(
-    "-csv",
+    "--csv",
     required=True,
     multiple=True,
     help=(
@@ -142,16 +141,6 @@ def import_targets(
     """
     settings = context.obj["settings"]
     server = settings.get("SERVER")
-    token = settings.get("TOKEN")
-    if server and token:
-        token_access_check = is_logged(server, token)
-        if token_access_check.state is False:
-            raise click.ClickException(
-                f"{str(token_access_check.data)}"
-                "\n\nTry re-login: 'Repository Service for TUF admin login'"
-            )
-    else:
-        raise click.ClickException("Login first. Run 'rstuf admin login'")
 
     headers = get_headers(settings)
 
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -95,7 +95,10 @@ def _import_csv_to_rstuf(
     "-csv",
     required=True,
     multiple=True,
-    help="CSV file to import. Multiple -csv parameters are allowed.",
+    help=(
+        "CSV file to import. Multiple -csv parameters are allowed. "
+        "See rstuf CLI guide for more details."
+    ),
 )
 @click.option(
     "--skip-publish-targets",
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -1,0 +1,168 @@
+import json
+import os
+from datetime import datetime
+from typing import Any, Dict, List
+
+from rich.console import Console
+from sqlalchemy import Connection, MetaData, Table, create_engine
+from sqlalchemy.exc import IntegrityError
+
+from repository_service_tuf.cli import click
+from repository_service_tuf.cli.admin import admin
+from repository_service_tuf.helpers.api_client import (
+    URL,
+    Methods,
+    get_headers,
+    is_logged,
+    request_server,
+    task_status,
+)
+from repository_service_tuf.helpers.tuf import Metadata, SuccinctRoles
+
+console = Console()
+
+
+def _check_csv_files(csv_files: List[str]):
+    not_found_csv_files: List[str] = []
+    for csv_file in csv_files:
+        if not os.path.isfile(csv_file):
+            not_found_csv_files.append(csv_file)
+
+    if len(not_found_csv_files) > 0:
+        raise click.ClickException(
+            f"CSV file(s) not found: {(', ').join(not_found_csv_files)}"
+        )
+
+
+def _parse_csv_data(
+    csv_file: str, succinct_roles: SuccinctRoles
+) -> List[Dict[str, Any]]:
+    rstuf_db_data: List[Dict[str, Any]] = []
+    with open(csv_file, "r") as f:
+        for line in f:
+            rstuf_db_data.append(
+                {
+                    "path": line.split(";")[0],
+                    "info": {
+                        "length": int(line.split(";")[1]),
+                        "hashes": {line.split(";")[2]: line.split(";")[3]},
+                    },
+                    "rolename": succinct_roles.get_role_for_target(
+                        line.split(";")[0]
+                    ),
+                    "published": False,
+                    "action": "ADD",
+                    "last_update": datetime.now(),
+                }
+            )
+
+    return rstuf_db_data
+
+
+def _import_csv_to_rstuf(
+    db_client: Connection,
+    rstuf_table: Table,
+    csv_files: List[str],
+    succinct_roles: SuccinctRoles,
+) -> None:
+    for csv_file in csv_files:
+        console.print(f"Import status: Loading data from {csv_file}")
+        rstuf_db_data = _parse_csv_data(csv_file, succinct_roles)
+        console.print(f"Import status: Importing {csv_file} data")
+        try:
+            db_client.execute(rstuf_table.insert(), rstuf_db_data)
+        except IntegrityError:
+            raise click.ClickException(
+                "Import status: ABORTED due duplicated targets. "
+                "CSV files must to have unique targets (path). "
+                "No data added to RSTUF DB."
+            )
+        console.print(f"Import status: {csv_file} imported")
+
+
+@admin.command()
+@click.option(
+    "-metadata-url",
+    required=True,
+    help="RSTUF Metadata URL i.e.: http://127.0.0.1 .",
+)
+@click.option(
+    "-db-uri",
+    required=True,
+    help="RSTUF DB URI. i.e.: postgresql://postgres:secret@127.0.0.1:5433",
+)
+@click.option(
+    "-csv",
+    required=True,
+    multiple=True,
+    help="CSV file to import. Multiple -csv parameters are allowed.",
+)
+@click.option(
+    "--skip-publish-targets",
+    is_flag=True,
+    help="Skip publish targets process in TUF Metadata.",
+)
+@click.pass_context
+def import_targets(context, metadata_url, db_uri, csv, skip_publish_targets):
+    """
+    Import targets to RSTUF from exported CSV file.
+    """
+    settings = context.obj["settings"]
+    server = settings.get("SERVER")
+    token = settings.get("TOKEN")
+    if server and token:
+        token_access_check = is_logged(server, token)
+        if token_access_check.state is False:
+            raise click.ClickException(
+                f"{str(token_access_check.data)}"
+                "\n\nTry re-login: 'Repository Service for TUF admin login'"
+            )
+    else:
+        raise click.ClickException("Login first. Run 'rstuf admin login'")
+
+    headers = get_headers(settings)
+
+    response = request_server(metadata_url, "1.bin.json", Methods.get)
+    if response.status_code == 404:
+        raise click.ClickException("RSTUF Metadata Targets not found.")
+
+    # load all required infrastructure
+    json_data = json.loads(response.text)
+    targets = Metadata.from_dict(json_data)
+    succinct_roles = targets.signed.delegations.succinct_roles
+    engine = create_engine(f"{db_uri}")
+    db_metadata = MetaData()
+    db_client = engine.connect()
+    rstuf_table = Table("rstuf_targets", db_metadata, autoload_with=engine)
+
+    # validate if the CSV files are accessible
+    _check_csv_files(csv)
+    # import all CSV file(s) data to RSTUF DB without commiting
+    _import_csv_to_rstuf(db_client, rstuf_table, csv, succinct_roles)
+
+    # commit data into RSTUF DB
+    console.print("Import status: Commiting all data to the RSTUF database")
+    db_client.commit()
+    console.print("Import status: All data imported to RSTUF DB")
+
+    if skip_publish_targets:
+        console.print(
+            "Import status: Finshed. "
+            "Not targets published (`--skip-publish-targets`)"
+        )
+    else:
+        console.print("Import status: Submitting action publish targets")
+        publish_targets = request_server(
+            server, URL.publish_targets.value, Methods.post, headers=headers
+        )
+        if publish_targets.status_code != 202:
+            raise click.ClickException(
+                f"Failed to publish targets. {publish_targets.text}"
+            )
+        task_id = publish_targets.json()["data"]["task_id"]
+        console.print(f"Import status: Publish targets task id is {task_id}")
+
+        # monitor task status
+        result = task_status(task_id, server, headers, "Import status: task ")
+        if result is not None:
+            console.print("Import status: [green]Finished.[/]")
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -101,8 +101,7 @@ def _get_succinct_roles(metadata_url: str) -> SuccinctRoles:
     if targets_delegations.succinct_roles is None:
         raise click.ClickException("Failed to get Targets succinct roles")
 
-    succinct_roles: SuccinctRoles = targets_delegations.succinct_roles
-    return succinct_roles
+    return targets_delegations.succinct_roles
 
 
 @admin.command()
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -119,7 +119,7 @@ def _get_succinct_roles(metadata_url: str) -> SuccinctRoles:
     required=True,
     multiple=True,
     help=(
-        "CSV file to import. Multiple -csv parameters are allowed. "
+        "CSV file to import. Multiple --csv parameters are allowed. "
         "See rstuf CLI guide for more details."
     ),
 )
--- a/repository_service_tuf/helpers/api_client.py
+++ b/repository_service_tuf/helpers/api_client.py
@@ -1,20 +1,25 @@
 # SPDX-FileCopyrightText: 2022-2023 VMware Inc
 #
 # SPDX-License-Identifier: MIT
-
+import time
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Dict, Optional
 
 import requests
+from requests.exceptions import ConnectionError
+from rich.console import Console
 
 from repository_service_tuf.cli import click
+
+console = Console()
 
 
 class URL(Enum):
     token = "api/v1/token/"  # nosec bandit: not hard coded password.
     bootstrap = "api/v1/bootstrap/"
     task = "api/v1/task/?task_id="
+    publish_targets = "api/v1/targets/publish/"
 
 
 class Methods(Enum):
@@ -50,7 +55,7 @@ def request_server(
         else:
             raise ValueError("Internal Error. Invalid HTTP/S Method.")
 
-    except requests.exceptions.ConnectionError:
+    except ConnectionError:
         raise click.ClickException(f"Failed to connect to {server}")
 
     return response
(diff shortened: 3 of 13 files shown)


kairoaraujo:
> This feature implements #39, right?

It partially implements #39.
The #39 is about adding targets `add-targets`. This feature should be an interface for the REST API 
It ended up a bit confusing.

Should we create a specific issue for it? Should I also implement add-targets to this PR?

> Because in your comment here [vmware/repository-service-tuf#188 (comment)](https://github.com/vmware/repository-service-tuf/issues/188#issuecomment-1401815310) you described all of the steps required to finish #188 and two remain: one is documentation and the other is importing targets in the CLI.

What I mention is that "the feature is described in vmware/repository-service-tuf#188" and on the BDD.

MVrachev:
> > This feature implements #39, right?
> 
> It partially implements #39. The #39 is about adding targets `add-targets`. This feature should be an interface for the REST API It ended up a bit confusing.

I see what you did now.
I was confused about the state of #39 as in your comment https://github.com/vmware/repository-service-tuf-cli/issues/39#issuecomment-1412124948 you laid down only two todo items for #39, but it seems that there are more.


> Should we create a specific issue for it? Should I also implement add-targets to this PR?

I don't think it will be necerry to create a new issue now when we have a pr, but could you update https://github.com/vmware/repository-service-tuf-cli/issues/39#issuecomment-1412124948 comment with the remaining steps (mentioning this step of importing targets) and their status?

> > Because in your comment here [vmware/repository-service-tuf#188 (comment)](https://github.com/vmware/repository-service-tuf/issues/188#issuecomment-1401815310) you described all of the steps required to finish #188 and two remain: one is documentation and the other is importing targets in the CLI.
> 
> What I mention is that "the feature is described in [vmware/repository-service-tuf#188](https://github.com/vmware/repository-service-tuf/issues/188)" and on the BDD.

Yes, got it.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+) -> None:
+    for csv_file in csv_files:
+        console.print(f"Import status: Loading data from {csv_file}")
+        rstuf_db_data = _parse_csv_data(csv_file, succinct_roles)
+        console.print(f"Import status: Importing {csv_file} data")
+        try:
+            db_client.execute(rstuf_table.insert(), rstuf_db_data)
+        except IntegrityError:
+            raise click.ClickException(
+                "Import status: ABORTED due duplicated targets. "
```
MVrachev, line 76: We are sure this error will be thrown only when there are duplicating targets.
We have parsed the csv file already, so all format errors should be caught already, but any possible problem?
I am asking because the error message is really specific and for `IntegrityError` in the `sqlalchemy` documentation is [written](https://docs.sqlalchemy.org/en/14/errors.html#integrityerror):
```
IntegrityError
Exception raised when the relational integrity of the database is affected, e.g. a foreign key check fails.

This error is a [DBAPI Error](https://docs.sqlalchemy.org/en/14/errors.html#error-dbapi) and originates from the database driver (DBAPI), not SQLAlchemy itself.
```

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+@click.option(
+    "-csv",
+    required=True,
+    multiple=True,
+    help="CSV file to import. Multiple -csv parameters are allowed.",
+)
+@click.option(
+    "--skip-publish-targets",
+    is_flag=True,
+    help="Skip publish targets process in TUF Metadata.",
```
MVrachev, line 103: Shouldn't be:
``` suggestion
    help="Skip publishing targets in TUF Metadata.",
```

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+)
+@click.pass_context
+def import_targets(context, metadata_url, db_uri, csv, skip_publish_targets):
```
MVrachev, line 106: Can we annotate those types?
Or we can't mention with `click` what types do we expect for our parameters?

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+
+    headers = get_headers(settings)
+
+    response = request_server(metadata_url, "1.bin.json", Methods.get)
+    if response.status_code == 404:
+        raise click.ClickException("RSTUF Metadata Targets not found.")
+
+    # load all required infrastructure
+    json_data = json.loads(response.text)
+    targets = Metadata.from_dict(json_data)
```
MVrachev, line 131: You can use insted the `JSONDeserializer()` from `python-tuf`: https://github.com/theupdateframework/python-tuf/blob/2840f629b07409b610c8943a2614272f0ff02bbc/tuf/api/serialization/json.py#L30

Also, because you are relying on a specific python-tuf `Signed` type it will be good to add an annotation for `targets` like:
`targets: Metadata[Targets] = Metadata.from_dict(json_data)`

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+    succinct_roles = targets.signed.delegations.succinct_roles
+    engine = create_engine(f"{db_uri}")
+    db_metadata = MetaData()
+    db_client = engine.connect()
+    rstuf_table = Table("rstuf_targets", db_metadata, autoload_with=engine)
+
+    # validate if the CSV files are accessible
+    _check_csv_files(csv)
```
MVrachev, line 139: The name of the `csv` argument is a little confusing and given there is no type annotation on this function even more.
Is this a list of `csv` file names as the `_check_csv_files()` function expects or it's one?
What will happen if only one `csv` argument is given?
What if many are given?

repository_service_tuf/helpers/api_client.py
```diff
@@ -1,20 +1,25 @@
...
 from repository_service_tuf.cli import click
 
+console = Console()
+
 
 class URL(Enum):
     token = "api/v1/token/"  # nosec bandit: not hard coded password.
     bootstrap = "api/v1/bootstrap/"
     task = "api/v1/task/?task_id="
+    publish_targets = "api/v1/targets/publish/"
```
MVrachev, line 22: From what I know the enum types should be named as constants with all upper case.
This is also supported by the python docs on enums:
https://docs.python.org/3/library/enum.html
Do you think it's worth discussing in another issue?

MVrachev (commented):
I reviewed the core code without the tests yet.
I have too many questions/comments on it.

Overall the work is big. It's visible it caused you a great amount of time an effort.

Besides my comments: do you plan on adding a test file with tests specifically testing
`repository_service_tuf/cli/admin/import_targets.py`?

Pipfile
```diff
@@ -13,6 +13,8 @@ requests = "*"
 tuf = "==2.0.0"
 dynaconf = {extras = ["ini"], version = "*"}
 isort = "*"
+sqlalchemy = "*"
+psycopg2 = "*"
```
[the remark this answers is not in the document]
kairoaraujo: A comment on Pipefile will be removed, as the `pipenv` command will automatically generate this.
The best place will be to add it to the `pyproject.toml`
I will add this dependencies as optional dependencies once the user needs to install only to use the `import-targets` feature.
```toml
[project.optional-dependencies]
psycopg2 = ["psycopg2>=2.9.5"]  # required by import-targets sub-command
sqlalchemy = ["sqlalchemy>=2.0.1"]  # required by import-targets sub-command
```

repository_service_tuf/helpers/api_client.py
```diff
@@ -69,6 +74,80 @@ def is_logged(server: str, token: str):
...
+
+def get_headers(settings: Dict[str, str]) -> Dict[str, str]:
+    server = settings.get("SERVER")
+    token = settings.get("TOKEN")
+    if server and token:
+        token_access_check = is_logged(server, token)
+        if token_access_check.state is False:
+            raise click.ClickException(
+                f"{str(token_access_check.data)}"
+                "\n\nTry re-login: 'rstuf admin login'"
```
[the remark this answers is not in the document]
kairoaraujo: The amount of data can be large, and the login message can be "hidden" for the user.
It is more for the user experience.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+
+
+def _parse_csv_data(
+    csv_file: str, succinct_roles: SuccinctRoles
+) -> List[Dict[str, Any]]:
+    rstuf_db_data: List[Dict[str, Any]] = []
+    with open(csv_file, "r") as f:
+        for line in f:
+            rstuf_db_data.append(
```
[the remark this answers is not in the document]
kairoaraujo: That is a good catch.
Yes, I forgot to have a short `csv`. 
I think it must be with the help of the command line + in the documentation guide. For the user experience, they might not go to the BDD tests or even unit tests to look for an example.

repository_service_tuf/helpers/api_client.py
```diff
@@ -69,6 +74,80 @@ def is_logged(server: str, token: str):
...
+        if expired_admin is True:
+            raise click.ClickException(
+                "Token expired. Run 'rstuf admin login'"
+            )
+        else:
+            headers = {"Authorization": f"Bearer {token}"}
+            response = request_server(
+                server, URL.bootstrap.value, Methods.get, headers=headers
+            )
+            if response.status_code != 200:
```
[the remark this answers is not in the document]
kairoaraujo: Yes, the `import-targets` requires the bootstrap done. I will add the check.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,168 @@
...
+        token_access_check = is_logged(server, token)
+        if token_access_check.state is False:
+            raise click.ClickException(
+                f"{str(token_access_check.data)}"
+                "\n\nTry re-login: 'Repository Service for TUF admin login'"
+            )
+    else:
+        raise click.ClickException("Login first. Run 'rstuf admin login'")
+
+    headers = get_headers(settings)
```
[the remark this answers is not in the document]
kairoaraujo: We need to check the bootstrap before importing targets. Otherwise will not have TUF metadata to publish targets.

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: We validate only if the CSV file exists, and we parse it. We don't look for duplicates, and we don't want do it.

The reason is we can use the insert for that. Once it is about (`click.ClickException`), the insert is not committed to the DB. We fail without changing the DB.
If we add the check before, we need to implement the algorithm to check for duplicates, and the `insert()` will do it again once the DB uses it as *unique*

KAUTH:
I will have a look tomorrow and review as well :+1:

kairoaraujo:
@MVrachev, it is ready for another review.

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        fake_response = pretend.stub(
+            status_code=200,
+            json=pretend.call_recorder(lambda: {"data": {"bootstrap": True}}),
+        )
+        import_targets.request_server = pretend.call_recorder(
+            lambda *a, **kw: fake_response
+        )
+        import_targets._get_succinct_roles = pretend.call_recorder(
+            lambda *a: "fake_succinct_roles"
+        )
```
MVrachev, line 238: You forgot to assert the calls for `import_targets._get_succinct_roles` after invoking `import_targets._get_succinct_roles`.

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        assert result.exit_code == 1, result.output
+        assert "Error 500 Internal Error" in result.output
+        assert import_targets.is_logged.calls == [
+            pretend.call("fake-server", "test-token")
+        ]
+        assert import_targets.get_headers.calls == [
+            pretend.call(test_context["settings"])
+        ]
+        import_targets.request_server.calls == [
```
MVrachev, line 398:
``` suggestion
        assert import_targets.request_server.calls == [
```

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        assert (
+            "import-targets` requires bootstrap process done." in result.output
+        )
+        assert import_targets.is_logged.calls == [
+            pretend.call("fake-server", "test-token")
+        ]
+        assert import_targets.get_headers.calls == [
+            pretend.call(test_context["settings"])
+        ]
+        import_targets.request_server.calls == [
```
MVrachev, line 448:
``` suggestion
        assert import_targets.request_server.calls == [
```

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        result = client.invoke(
+            import_targets.import_targets, options, obj=test_context
+        )
+        assert result.exit_code == 0, result.output
+        assert "Finished." in result.output
+        assert "Not targets published" in result.output
+        assert import_targets.is_logged.calls == [
+            pretend.call("fake-server", "test-token")
+        ]
+        import_targets.request_server.calls == [
```
MVrachev, line 522:
``` suggestion
        assert import_targets.request_server.calls == [
```

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+            pretend.call(
+                "fake-server",
+                import_targets.URL.bootstrap.value,
+                import_targets.Methods.get,
+                headers="headers",
+            )
+        ]
+
+    def test_import_targets_skip_publish_targets(self, client, test_context):
```
MVrachev, line 457: Can we move this positive scenario closer to the rest of the positive scenarios?
What do you think?

MVrachev (changes requested):
Reviewed everything except the `test_api_client.py` file.
Still, I decided to submit my review as I have multiple questions and comments.

repository_service_tuf/helpers/api_client.py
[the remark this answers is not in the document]
kairoaraujo: `get_headers` returns only the headers.
The call used by `get_headers` to the bootstrap is to validate the token itself as the CLI token has all required scopes.

The bootstrap check from the ceremony and import-targets is the same data but has a different check.
- ceremony the expect a bootstrap not completed.
- import-targets expect a bootstrap completed.

repository_service_tuf/helpers/api_client.py
kairoaraujo: Issue created
- https://github.com/vmware/repository-service-tuf-cli/issues/197

tests/unit/cli/admin/test_ceremony.py
```diff
@@ -457,6 +457,7 @@ def test_ceremony_with_flag_bootstrap_forbidden(
                     "detail": "Unauthorized.",
                 }
             ),
+            text="Unauthorized."
```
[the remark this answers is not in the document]
kairoaraujo: It is related to `requests.Response`. The method `.json()` In the response is not always present. In case of error we should always parse the `Response.text`. 
It handles the errors better. I just fixed it in the ceremony.

repository_service_tuf/helpers/api_client.py
```diff
@@ -69,6 +74,94 @@ def is_logged(server: str, token: str):
             return Login(state=True, data=data)
 
     else:
-        click.ClickException(
-            f"Error {response.status_code} {response.json()['detail']}"
+        raise click.ClickException(
```
KAUTH, line 77: I believe `is_logged` does not work as expected in all cases.
A `GET api/v1/token/` request may also return a `200` response with a body of:
```
{
  "detail": {
    "error": "Failed to validate token"
  }
}
```
See [here](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/token.py#L210). In this case, `data` is `None` and ` data.get("expired")` will throw an `AttributeError`

I spotted it here, but if you validate it I can open a bug. (BTW quick fix: `data = response.json().get("data", {})`)

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,201 @@
...
+    token = settings.get("TOKEN")
+    if server and token:
+        token_access_check = is_logged(server, token)
+        if token_access_check.state is False:
+            raise click.ClickException(
+                f"{str(token_access_check.data)}"
+                "\n\nTry re-login: 'Repository Service for TUF admin login'"
+            )
+    else:
+        raise click.ClickException("Login first. Run 'rstuf admin login'")
```
KAUTH, line 154: This part still seems to me that IT is checked with `headers = get_headers(settings)` ([here](https://github.com/vmware/repository-service-tuf-cli/blob/1320eb54b127d35c438bb7f66761906b2661efee/repository_service_tuf/helpers/api_client.py#L85-L91)), why do we need it separately here?

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,201 @@
...
+
+    headers = get_headers(settings)
+
+    bs_response = request_server(
+        settings.SERVER, URL.bootstrap.value, Methods.get, headers=headers
+    )
+    if bs_response.status_code != 200:
+        raise click.ClickException(
+            f"Error {bs_response.status_code} {bs_response.text}"
+        )
```
KAUTH, line 164: Here we are doing the same exact request and response check as in [here](https://github.com/vmware/repository-service-tuf-cli/blob/1320eb54b127d35c438bb7f66761906b2661efee/repository_service_tuf/helpers/api_client.py#L100-L105).
I am thinking maybe instead of a `get_headers` method, do we want a `get_bootstrap` or `valid_bootstrap` which is the same as `get_headers` (meaning it does all the validation) and if it succeeds it returns the bootstrap response.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if targets_delegations.succinct_roles is None:
+        raise click.ClickException("Failed to get Targets succinct roles")
+
+    return targets_delegations.succinct_roles
+
+
+@admin.command()
+@click.option(
+    "-metadata-url",
```
KAUTH, line 109: Why here, in `-db-uri` and `-csv` do we not follow the same double dash pattern (`--`) that we do for `--skip-publish-targets`, since they are all multi-phrase options?

repository_service_tuf/helpers/api_client.py
```diff
@@ -69,6 +74,94 @@ def is_logged(server: str, token: str):
...
+    else:
+        raise click.ClickException("Login first. Run 'rstuf admin login'")
+
+    return headers
+
+
+def task_status(
+    task_id: str, server: str, headers: Dict[str, str], title: Optional[str]
+) -> Dict[Any, str]:
+    received_state = []
```
KAUTH, line 116: nit: `received_states`

repository_service_tuf/helpers/api_client.py
```diff
@@ -69,6 +74,94 @@ def is_logged(server: str, token: str):
...
+                )
+    else:
+        raise click.ClickException("Login first. Run 'rstuf admin login'")
+
+    return headers
+
+
+def task_status(
+    task_id: str, server: str, headers: Dict[str, str], title: Optional[str]
+) -> Dict[Any, str]:
```
KAUTH, line 115: nit: I think the type is `Dict[str, Any]`, taking into account https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/tasks.py#L21

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+
+        with pytest.raises(import_targets.click.ClickException) as err:
+            import_targets._get_succinct_roles("http://127.0.0.1/metadata")
+        assert "Failed to get Targets succinct roles" in str(err)
+
+
+class TestImportTargetsGroupCLI:
+    def test_import_targets(self, client, test_context):
+        test_context["settings"].SERVER = "fake-server"
+        test_context["settings"].TOKEN = "test-token"
```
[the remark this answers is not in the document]
kairoaraujo: `settings` is a `Dynaconf` object, which makes it a `LazySettings` object.

https://github.com/vmware/repository-service-tuf-cli/blob/913e1d8c1f773cb670e5abd99d88fce3f9642f79/tests/conftest.py#L16

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        assert "Finished." in result.output
+        assert import_targets.is_logged.calls == [
+            pretend.call("fake-server", "test-token")
+        ]
+        import_targets.request_server.calls == [
+            pretend.call(
+                "fake-server",
+                import_targets.URL.bootstrap.value,
+                import_targets.Methods.get,
+                headers="headers",
```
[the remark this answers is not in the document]
kairoaraujo: The pretend will record the call based on your function implementation. If the code implementation used a keyword to call it would be recorded as is.
https://github.com/vmware/repository-service-tuf-cli/blob/73b0c9c3e242c0c8faf10fcea5f32eabe1bfab4e/repository_service_tuf/cli/admin/import_targets.py#L160
To assert, you must use the same record. If you change the implementation not to use the keyword call, you will also catch it in the assets.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -0,0 +1,202 @@
...
+        console.print(f"Import status: {csv_file} imported")
+
+
+def _get_succinct_roles(metadata_url: str) -> SuccinctRoles:
+    response = request_server(metadata_url, "1.bin.json", Methods.get)
+    if response.status_code == 404:
+        raise click.ClickException("RSTUF Metadata Targets not found.")
+
+    json_data = json.loads(response.text)
+    targets: Metadata[Targets] = Metadata.from_dict(json_data)
```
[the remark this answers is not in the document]
kairoaraujo: > I spotted it here, but if you validate it I can open a bug. (BTW quick fix: `data = response.json().get("data", {})`)

That's is a good spot, I'm implementing the quick fix.

> On the other hand, we are always validating a token with the same token in the headers, so I assume if the token is invalid it won't be because it can't be decoded... so it might be fine this way. I will not delete it, but I believe you can ignore my comment.

Maybe we can have a specific issue for this part to implement or remove this check.
Can you file the issue giving the background?

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -0,0 +1,541 @@
...
+        fake_data = [
+            "path/file1;123;blake2b-256;hash1",
+            "path/file2;456;blake2b-256;hash2",
+        ]
+        fake_file_obj = pretend.stub(
+            __enter__=pretend.call_recorder(lambda: fake_data),
+            __exit__=pretend.call_recorder(lambda *a: None),
+            close=pretend.call_recorder(lambda: None),
+            read=pretend.call_recorder(lambda: fake_data),
+        )
```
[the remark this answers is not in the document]
kairoaraujo: I think is the same what @MVrachev [mentioned above](https://github.com/vmware/repository-service-tuf-cli/pull/195#discussion_r1104658018)

Actually, `get_headers` returns the headers used for specific calls (other endpoints) after checking if the token is still valid (`is_logged`) and returns the headers formatted.

We can have another helper in the `api_client` to retrieve the bootstrap status. Because the status will be used in different behaviors.

tests/unit/cli/admin/test_import_targets.py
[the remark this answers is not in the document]
kairoaraujo: It was defined here

https://github.com/vmware/repository-service-tuf/blob/c3d8ec77dfdc608a2aabd3a932d4206cc4ca8ab6/tests/features/targets/import_targets.feature#L15

I will change. We will need another PR change there.

tests/unit/cli/admin/test_import_targets.py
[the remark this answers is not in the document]
kairoaraujo: @MVrachev , @KAUTH 
I was simulating this change here, it will endup in a lot of changes.
I think would be better do it in a different PR, otherwise it will deviate from the initial PR goal.

repository_service_tuf/helpers/api_client.py
kairoaraujo: > I spotted it here, but if you validate it I can open a bug. (BTW quick fix: `data = response.json().get("data", {})`)

That's a good spot. I'm implementing the quick fix.

> On the other hand, we are always validating a token with the same token in the headers, so I assume if the token is invalid it won't be because it can't be decoded... so it might be fine this way. I will not delete it, but I believe you can ignore my comment.

Maybe we can have a specific issue for this part to implement or remove this check.
Can you file the issue giving the background?

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: I think is the same what @MVrachev [mentioned above](https://github.com/vmware/repository-service-tuf-cli/pull/195#discussion_r1104658018)

Actually, `get_headers` returns the headers used for specific calls (other endpoints) after checking if the token is still valid (`is_logged`) and returns the headers formatted.

We can have another helper in the `api_client` to retrieve the bootstrap status. Because the status will be used in different behaviors.

kairoaraujo: @MVrachev , @KAUTH
I was simulating this change here, it will endup in a lot of changes.
I think would be better do it in a different PR, otherwise it will deviate from the initial PR goal.

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: It was decided in the [Feature PR](https://github.com/vmware/repository-service-tuf/blob/feature/import-targets-to-db/tests/features/targets/import_targets.feature).
But I changed it.
I will open a PR to update the Feature file.

repository_service_tuf/cli/admin/import_targets.py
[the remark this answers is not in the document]
kairoaraujo: As we discussed, it gets the readability confused. So we decided to keep it as is.

repository_service_tuf/cli/admin/import_targets.py
KAUTH: > I was simulating this change here, it will endup in a lot of changes.
I think would be better do it in a different PR, otherwise it will deviate from the initial PR goal.

I am ok with that :+1: , there are indeed already a lot of changes

MVrachev: Do we want to discuss this in a separate issue then?
If so, I think there are two discussions that it's good to be linked:
- this one
- and the one I started previously: https://github.com/vmware/repository-service-tuf-cli/pull/195#discussion_r1101454006

repository_service_tuf/helpers/api_client.py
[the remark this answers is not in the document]
MVrachev: As I have mentioned in my comment here https://github.com/vmware/repository-service-tuf-cli/pull/195/commits/82b64ab4e26cebb93d298336f0c7fad987afd8ad#r1106678265 we can continue this discussion in a separate issue as there are already lots of changes.

tests/unit/cli/admin/test_import_targets.py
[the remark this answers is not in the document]
MVrachev: What do you want to change?

repository_service_tuf/helpers/api_client.py
MVrachev: @KAUTH when you open it could please link it here for future reference?

repository_service_tuf/cli/admin/import_targets.py
MVrachev: I just thought you used a single `-` when it's a parameter with a value and `--` when it's a flag.
Isn't that the style for CLI tools usually?

tests/unit/cli/admin/test_import_targets.py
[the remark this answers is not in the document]
kairoaraujo: This comment went to the wrong thread; it is about the `- `or `--` in the commands. All my comments yesterday went to the wrong threads, hiding them.

repository_service_tuf/cli/admin/import_targets.py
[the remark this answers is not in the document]
kairoaraujo: Issue opened
- https://github.com/vmware/repository-service-tuf-cli/issues/198

repository_service_tuf/helpers/api_client.py
kairoaraujo: Issue filed with coment referencing both discussions.

- https://github.com/vmware/repository-service-tuf-cli/issues/199

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: There are no *written* rules.
But what is more common is `-m/--multiple-word-parameters`. We can do it and make it easier to follow this rule in CLI. The short parameter we are not adding but adding in the future does not impact the current implementation. It works more as an alias.

repository_service_tuf/helpers/api_client.py
kairoaraujo: I opened this issue
- https://github.com/vmware/repository-service-tuf-cli/issues/198

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: I opened this issue with both references.
- https://github.com/vmware/repository-service-tuf-cli/issues/199

tests/unit/helpers/test_api_client.py
```diff
@@ -2,57 +2,386 @@
...
+                "http://server",
+                "api/v1/token/?token=fake_token",
+                api_client.Methods.get,
+                headers={"Authorization": "Bearer fake_token"},
+            )
+        ]
+
+    def test_get_headers(self):
+        api_client.is_logged = pretend.call_recorder(
+            lambda *a, **kw: api_client.Login(
```
MVrachev, line 135: You don't need to use `**kw` in this lambda.

tests/unit/helpers/test_api_client.py
```diff
@@ -2,57 +2,386 @@
...
+
+    def test_get_headers_never_logged(self):
+        with pytest.raises(api_client.click.ClickException) as err:
+            api_client.get_headers({})
+
+        assert "Login first. Run 'rstuf admin login'" in str(err)
+
+    def test_get_headers_is_logged_state_false(self):
+        api_client.is_logged = pretend.call_recorder(
+            lambda *a, **kw: api_client.Login(
```
MVrachev, line 168: You don't need to use `**kw` in this lambda.

tests/unit/helpers/test_api_client.py
```diff
@@ -2,57 +2,386 @@
...
+            )
+
+        assert "re-login" in str(err)
+        assert api_client.is_logged.calls == [
+            pretend.call("http://server", "fake_token")
+        ]
+
+    def test_get_headers_is_logged_state_true_expired_token(self):
+        api_client.is_logged = pretend.call_recorder(
+            lambda *a, **kw: api_client.Login(
```
MVrachev, line 185: You don't need to use `**kw` in this lambda.

tests/unit/helpers/test_api_client.py
```diff
@@ -2,57 +2,386 @@
...
+            )
+
+        assert "The token has expired" in str(err)
+        assert api_client.is_logged.calls == [
+            pretend.call("http://server", "fake_token")
+        ]
+
+    def test_get_headers_unexpected_error(self):
+        api_client.is_logged = pretend.call_recorder(
+            lambda *a, **kw: api_client.Login(
```
MVrachev, line 202: You don't need to use `**kw` in this lambda.

tests/unit/helpers/test_api_client.py
```diff
@@ -2,57 +2,386 @@
...
+            )
+
+        assert "Unexpected error" in str(err)
+        assert api_client.is_logged.calls == [
+            pretend.call("http://server", "fake_token")
+        ]
+
+    def test_task_status(self):
+        fake_json = Mock()
+        fake_json.side_effect = [
```
MVrachev, line 221: I am unfamiliar with `Mock()` objects.
What does `side_effect` do and why you decided to use a `Mock` object?

kairoaraujo: Unfortunately, pretend can't make multiple different mocked retuvariousmultiple calls. I'm even thinking of submitting a PR to implement it there.

`Mock()` creates a mocked object and the `side_effect` gives a sequence of mocked returns (from a List) or even an `Exception`. [See the documentation.](https://docs.python.org/3/library/unittest.mock.html#quick-guide)
As the implementation does, `def task_status` calls multiple times to the REST API tasks. 

https://github.com/vmware/repository-service-tuf-cli/blob/353f44b79060546a4a976080ce940e11af5ba92a/repository_service_tuf/helpers/api_client.py#L117-L120

I simulate multiple returns with different responses, as it is in the actual call with different statuses. I test the logic unit from the implementation.

repository_service_tuf/cli/admin/import_targets.py
MVrachev: I am okay with using `-` for short one-word parameters and `--` for multi-word parameters.
By the way: if we decide to make changes and to use `--` for any of the parameters we need to update the documentation in the `docs/source/guide/index.rst` as well.

tests/unit/helpers/test_api_client.py
MVrachev: That's an interesting use case. Thanks for sharing that info, it will be useful.

MVrachev (changes requested):
I see only one place that requires change and I have described it in my comment:
https://github.com/vmware/repository-service-tuf-cli/pull/195#discussion_r1107080897

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: Updated.

MVrachev (approved):
Great job @kairoaraujo!
LGTM from me, but will wait for another day in hope that @KAUTH can review it again.

KAUTH (approved):
I had a look at the tests as well, LGTM.
A big PR, nice work guys!

MVrachev:
With two approvals how could you not merge this :))

PR #210, opened 2023-02-24
Add marks for including accross docs

kairoaraujo:
This is to be used by documentation in Guide.
It doesn't add extra docs for users.

- Used by https://github.com/vmware/repository-service-tuf/pull/254

### DIFF
--- a/docs/source/guide/index.rst
+++ b/docs/source/guide/index.rst
@@ -32,6 +32,8 @@ Using pip:
 Administration (``admin``)
 ==========================
 
+.. rstuf-cli-admin
+
 It executes administrative commands to the Repository Service for TUF.
 
 .. code:: shell
@@ -54,6 +56,8 @@ It executes administrative commands to the Repository Service for TUF.
 
 Login to Server (``login``)
 ---------------------------
+
+.. rstuf-cli-admin-login
 
 This command will log in to Repository Service for TUF and give you a token to run other commands
 such as Ceremony, Token Generation, etc.
@@ -81,6 +85,8 @@ such as Ceremony, Token Generation, etc.
 
 Ceremony (``ceremony``)
 -----------------------
+
+.. rstuf-cli-admin-ceremony
 
 The Repository Service for TUF Metadata uses the following Roles: ``root``, ``timestamp``,
 ``snapshot``, ``targets``, ``bin``, and ``bins`` to build the Repository
@@ -367,6 +373,8 @@ Using another computer with access to ``repository-service-tuf-api``
 Token (``token``)
 -----------------
 
+.. rstuf-cli-admin-token
+
 Token Management
 
 .. code:: shell
@@ -387,6 +395,9 @@ Token Management
 
 ``generate``
 ............
+
+.. rstuf-cli-admin-token-generate
+
 Generate tokens to use in integrations.
 
 .. code:: shell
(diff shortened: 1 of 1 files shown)


MVrachev (commented):
When you use import a part of this document the way you did in https://github.com/vmware/repository-service-tuf/pull/254:

```
.. include:: ../repository-service-tuf-cli/index.rst
    :start-after: rstuf-cli-admin
```
this will import everything after the `rstuf-cli-admin` mark and until it finds a new mark, correct?

kairoaraujo:
> When you use import a part of this document the way you did in [vmware/repository-service-tuf#254](https://github.com/vmware/repository-service-tuf/pull/254):
> 
> ```
> .. include:: ../repository-service-tuf-cli/index.rst
>     :start-after: rstuf-cli-admin
> ```
> 
> this will import everything after the `rstuf-cli-admin` mark and until it finds a new mark, correct?

You can use the options `start-after`, `end-before` for the `include` directive
See https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment

MVrachev:
> > When you use import a part of this document the way you did in [vmware/repository-service-tuf#254](https://github.com/vmware/repository-service-tuf/pull/254):
> > ```
> > .. include:: ../repository-service-tuf-cli/index.rst
> >     :start-after: rstuf-cli-admin
> > ```

> > this will import everything after the `rstuf-cli-admin` mark and until it finds a new mark, correct?
> 
> You can use the options `start-after`, `end-before` for the `include` directive See https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment

I read it and it's clear why you added `start-after`.
My question is don't we need to also and `end-before` or something like this in order to not import more than we want we do for example:
```
.. include:: ../repository-service-tuf-cli/index.rst
    :start-after: rstuf-cli-admin-import-targets
```
I don't see an example where we have used this to check.

kairoaraujo:
> I read it and it's clear why you added `start-after`. My question is don't we need to also and `end-before` or something like this in order to not import more than we want we do for example:
> 
> ```
> .. include:: ../repository-service-tuf-cli/index.rst
>     :start-after: rstuf-cli-admin-import-targets
> ```
> 
> I don't see an example where we have used this to check.

We don't need to add `end-before` because the `rstuf-cli-admin-import-targets` is the last part of the `/repository-service-tuf-cli/index.rst`.

But you can use it as well.

MVrachev:
Yes, I agree for `rstuf-cli-admin-import-targets`, but don't we expect we would include the other sections of this document as well?
Why then we add marks for the rest of the document if we don't expect them to be imported?

kairoaraujo:
> Yes, I agree for `rstuf-cli-admin-import-targets`, but don't we expect we would include the other sections of this document as well?
> Also, we could add a next section after that one tomorrow and we should not forget to add `end-before` then.

We don't know yet. We could add `rstuf-cli-end-of-file`

> Why then we add marks for the rest of the document if we don't expect them to be imported?

We add because we use the `` `` notation in the title, which makes the built-in title marks not work properly.
But, we add for all also to be consistent in this file.

MVrachev (approved):
Approving.
If in the future we see there are any problems with how we import or mark our `.rst` files we can update it accordingly.

Issue #223, opened 2023-03-09
Docs: update the documentation with the new CLI  


kairoaraujo:
### What do you want to share with us?

**! This is related to the feature/roles-simplification branch !**

This task is about updating the `docs/source/guide/` for the CLI
As we implement the new 'interface' for user interaction, we need to give the correct screens in the Guide.

We need an update: https://github.com/vmware/repository-service-tuf-cli/blob/feature/roles-simplification/docs/source/guide/index.rst

### References

 - https://github.com/vmware/repository-service-tuf/issues/207

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #228, opened 2023-03-21
Update feature/roles-simplification with latest main

kairoaraujo:
This PR updates the features/roles-simplification branch with the latest commits
- This is required for #223 
- There is a conflict that will be solved in a separate commit
- **Do not squash & merge**


kairoaraujo:
I will prepare a better PR

PR #232, opened 2023-03-22
Bootstrap ceremony: remove metadata creation

### LINKED ISSUE #177: Simplify the bootstrap ceremony
### What is the task about?

Related to the bigger task https://github.com/vmware/repository-service-tuf/issues/23 we have decided to simplify the bootstrap process https://github.com/vmware/repository-service-tuf/issues/208.

What should be done in the CLI `cli.admin.ceremony` file is to:
- remove the snapshot, timestamp, and bINS keys process loading
- remove target hash bins, snapshot, and timestamp metadata creation

We can remove those roles as they will be created in the worker and is not required to send them through `payload.json`.

### Parent feature

Related to the bigger task https://github.com/vmware/repository-service-tuf/issues/208

### References

_No response_

### Code of Conduct

MVrachev:
The main idea of this pr is to remove the creation of all metadata files in the bootstrap ceremony except for the root role.
We have decided that given that all roles except root share the same online key,
then there is no need to generate metadata for timestamp, snapshot, targets, 
and all delegated bins.
Instead, we can generate metadata only for root and inside the RSTUF worker
we would create the rest of the metadata files as we would have
access to the online key.

Before doing the main idea of this pr I decided to do some "house cleaning".
Meaning I decided to add tests for every help function inside
`repository_service_tuf/helpers/tuf.py` module.
This led to the need for the `TUFManagement` class and splitting the bigger
functions into even more helper functions for easier testing.

Additionally, I found a few places which can benefit from better validation.

Finally, while linting and checking that everything is accepted by our linters I
found out that `mypy` causes maaany warnings on our tests for things that
are required like mocking or pretending test functions.
I decided to remove `mypy` linting for the whole tests folder.
If there are other opinions I would like to hear them.
cc: @KAUTH 

Close #177


kairoaraujo:
After loading the online keys, when loading the root key(s) it asks for **bins** key instead root role.
Can you check?

```shell
                                           The Online Key                                           

The online key is the same one provided to the Repository Service for TUF Workers (RSTUF Worker).   
This key is responsible for signing the snapshot, timestamp, targets, and delegated targets (hash   
bin) roles.                                                                                         

The RSTUF Worker uses this key during the process of managing the metadata.                         

Note: the ceremony process won't show any password or key content.                                  

Choose 1/1 ONLINE key type [ed25519/ecdsa/rsa] (ed25519): 
Enter 1/1 the ONLINE`s private key path: tests/files/online.key
Enter 1/1 the ONLINE`s private key password: 
✅ Key 1/1 Verified
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                                      STEP 3: Load Root Keys                                      ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛


                                             Root Keys                                              

The keys must have a password, and the file must be accessible.                                     

Depending on the organization, each key has an owner, and each owner should insert the password     
personally.                                                                                         

Note: the ceremony process won't show any password or key content.                                  

Choose 1/2 bins key type [ed25519/ecdsa/rsa] (ed25519): 
Enter 1/2 the bins`s private key path:
```

kairoaraujo (changes requested):
Thank you for the great work @MVrachev. 🎉 

A general comment is that I considered removing `bump_version` and `bump_expirey` to simplify the initialization.
But it will be useful for Key Rotation.

MVrachev:
I decided that `bump_version()` can be removed.
Also, by following your comment https://github.com/vmware/repository-service-tuf-cli/pull/232#discussion_r1147793756 and not using versioning for role names I noticed that we no longer need `TUFManagement._load()`.

MVrachev:
I had to create a new branch from `main`, cherry-pick all changes from those 15 commits, rename the branch to my current `simplify-bootstrap-ceremony` and force push. 
Otherwise, when I just changed the base an additional 12 commits showed.

kairoaraujo (approved):
We need merge first:
- https://github.com/vmware/repository-service-tuf-worker/pull/243
- https://github.com/vmware/repository-service-tuf-api/pull/301

kairoaraujo (approved):

PR #253, opened 2023-04-12
Add parameter `--no-auth` to the CLI

### LINKED ISSUE #251: Add param `--no-auth` to skip the RSTUF API authentication
The RSTUF API authentication will be an optional functionality, where the RSTUF admin can choose on/off (on by default).

The RSTUF CLI will require an option to skip the authentication when connecting the API.
This makes RSTUF CLI compatible with https://github.com/vmware/repository-service-tuf-api/issues/282

kairoaraujo:
This adds the `rstuf --no-auth` parameter.

This parameter skips authentication checks/routines when communicating with RSTUF API. This makes RSTUF CLI compatible when the user disables RSTUF API built-in authentication.

Added a small refactoring about using the rich console and shared calls in the helper.
Added refactoring for the CLI init test file.

Closes #251

### DIFF
--- a/repository_service_tuf/cli/__init__.py
+++ b/repository_service_tuf/cli/__init__.py
@@ -9,10 +9,13 @@ import re
 from pathlib import Path
 
 import rich_click as click  # type: ignore
+from rich.console import Console
+from rich.panel import Panel
 
 from repository_service_tuf import Dynaconf
 from repository_service_tuf.__version__ import version
 
+console = Console()
 HOME = str(Path.home())
 
 # attempt to find the program name from pyproject.toml or give a default
@@ -39,17 +42,37 @@ except FileNotFoundError:
     help="Repository Service for TUF config file",
     required=False,
 )
+@click.option(
+    "--no-auth",
+    "auth",
+    help="Skips use RSTUF built-in Authentication",
+    is_flag=True,
+    default=True,
+    required=False,
+)
 # adds the --version parameter
 @click.version_option(prog_name=prog_name, version=version)
 @click.pass_context
-def rstuf(context, config):
+def rstuf(context, config, auth):
     """
     Repository Service for TUF Command Line Interface (CLI).
     """
     context.obj = {
         "settings": Dynaconf(settings_files=[config]),
         "config": config,
+        "auth": auth,
     }
+    settings = context.obj["settings"]
+    if auth is False:
+        console.print(
+            Panel(
+                "[white]Skipping RSTUF authentication (--no-auth)[/]",
+                title="[green]Info[/]",
+                title_align="left",
+                style="green",
+            )
+        )
+    settings.AUTH = auth
 
 
 # Register all command groups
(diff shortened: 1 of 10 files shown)


tests/unit/helpers/test_api_client.py
```diff
@@ -157,57 +170,68 @@ def test_get_headers(self):
             )
         ]
 
+    def test_get_headers_no_auth(self, test_context):
+        test_context["settings"].SERVER = "http://server"
+        test_context["settings"].TOKEN = "fake_token"
+        test_context["settings"].AUTH = False
```
MVrachev, line 176: I am not sure why you add settings here as they won't be accessed.

MVrachev (changes requested):
I noticed that if you use `rstuf --no-auth admin --help`
you will see two commands related to authentication: `login` and `token`.
Is there an easy way to hide them from the help menu if `--no-auth` is used?

repository_service_tuf/cli/admin/login.py
```diff
@@ -117,19 +114,24 @@ def login(context, force, server_, user_, password_, expires_):
     Login to Repository Service for TUF (API).
     """
     settings = context.obj.get("settings")
-    server = settings.get("SERVER")
-    token = settings.get("TOKEN")
 
-    if force is False and server is not None and token is not None:
-        response = is_logged(server, token)
+    if settings.get("AUTH") is False:
+        return None
```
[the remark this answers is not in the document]
kairoaraujo: I found way to disable login when `--no-auth`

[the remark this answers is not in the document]
MVrachev: Yep, I confirm that.

repository_service_tuf/cli/__init__.py
```diff
@@ -6,13 +6,17 @@
 import os
 import pkgutil
 import re
+import sys
 from pathlib import Path
 
 import rich_click as click  # type: ignore
+from rich.console import Console
```
MVrachev, line 13: I noticed in other files you remove those lines:
```
from rich.console import Console
console = Console()
```

and instead use an instance from `rich` directly like this:
```
from repository_service_tuf.cli import console
```
What is the difference between the two and is there a reason why here is different?

kairoaraujo:
> Now the only thing remaining is a `server` option which will be used only when `--no-auth` is used.
> This is required in order to be able to upload the bootstrap payload online.

It was implemented.
Check out the ceremony

repository_service_tuf/cli/__init__.py
kairoaraujo: No difference, as the `cli.__init__.py` is used by them. I decided to import it from the existing instance instead creating multiple console instances for each command file.

kairoaraujo:
> @kairoaraujo why the `--upload-server` is hidden?
> That's why I thought it was not yet implemented. I didn't see it in the help menu.
> Maybe you want to escape the confusion and show it only when there is an error like that:
> image

Correct.
I did that because it can be confusing and lead to a misusage by users without `--no-auth`.

MVrachev (approved):

PR #259, opened 2023-04-13
fix: option --no-auth breaking --upload with auth

kairoaraujo:
fix: option --no-auth breaking --upload with auth
    
    This fixes the logic behind `--no-auth` and `--upload` requiring
    `--upload-server`.
    
    When the user doesn't use `--no-auth`, but uses `--upload` the current
    logic was overwitting the server defined by `rstuf admin login`.
    
    Added tests to prevent regression

### DIFF
--- a/tests/unit/cli/admin/test_ceremony.py
+++ b/tests/unit/cli/admin/test_ceremony.py
@@ -708,6 +708,7 @@ class TestCeremonyOptions:
         )
         ceremony.task_status = pretend.call_recorder(lambda *a: None)
 
+        test_context["settings"].SERVER = "http://server"
         test_result = client.invoke(
             ceremony.ceremony,
             ["--bootstrap", "--upload"],
@@ -734,6 +735,8 @@ class TestCeremonyOptions:
                 "fake_task_id", test_context["settings"], "Bootstrap status: "
             )
         ]
+        # test regression https://github.com/vmware/repository-service-tuf-cli/pull/259  # noqa
+        assert test_context["settings"].SERVER is not None
 
     def test_ceremony_option_bootstrap_upload_no_auth(
         self, client, test_context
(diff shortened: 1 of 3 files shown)


tests/unit/cli/admin/test_ceremony.py
```diff
@@ -735,6 +735,67 @@ def test_ceremony_option_bootstrap_upload(self, client, test_context):
...
+        test_result = client.invoke(
+            ceremony.ceremony,
+            ["--bootstrap", "--upload"],
+            input=None,
+            obj=test_context,
+        )
+
+        assert test_result.exit_code == 1, test_result.output
+        assert (
+            "Requires '--upload-server' when using '--no-auth'"
```
MVrachev, line 795: I think this test is not a regression test.
This error message is before the check for `upload_server` which you actually change.
I think in order to make a real regression test you need to add authorization and verify that the `settings.SERVER` will be equal to `--upload-server` which is not provided, so it will be `None` instead of being the real value stored inside `rstuf.ini`.

MVrachev (changes requested):
Great catch, but I think we don't have a regression test yet.

tests/unit/cli/admin/test_ceremony.py
kairoaraujo: Thank you for the observation!
I added the check in the last commit.

In case of regression, it raises this error
    
```shell
        # test regression https://github.com/vmware/repository-service-tuf-cli/pull/259  # noqa
>       assert test_context["settings"].SERVER is not None
E       assert None is not None
E        +  where None = <dynaconf.base.LazySettings object at 0x10e08f0a0>.SERVER

tests/unit/cli/admin/test_ceremony.py:739: AssertionError
```

MVrachev (approved):

PR #266, opened 2023-04-17
fix: FT triggered from RSTUF CLI

kairoaraujo:
- It adds the folder with keys following the same standard as RSTUF API and Worker: `tests/files/key_storage`
- Add a small disclaimer to not use these keys on deploy


MVrachev (approved):
Good catch!

PR #289, opened 2023-05-18
Change CLI to not use built-in auth by default

kairoaraujo:
As implemented the task which makes the RSTUF API
Authentication/Authorization disabled by default
(https://github.com/repository-service-tuf/repository-service-tuf-api/issues/282) the default behavior of RSTUF CLI to not use auth by default.

If the user wants to use the CLI to the API with built-in authentication, it requires the parameter `--auth`.

`rstuf --auth admin`


KAUTH (changes requested):
Everything looks good to me (the very good test coverage really helps with these changes!). I believe we just need to rebuild the docs, see `docs/source/guide/index.rst`, line 27 still refers to `no-auth` option.

kairoaraujo:
> Everything looks good to me (the very good test coverage really helps with these changes!). I believe we just need to rebuild the docs, see `docs/source/guide/index.rst`, line 27 still refers to `no-auth` option.

Nice catch @KAUTH, just fixed it.

KAUTH (approved):
LGTM

PR #304, opened 2023-06-22
Fix import targets and simplify succinct roles

### LINKED ISSUE #297: Bug: `rstuf import-targets` doesn't work
### What happened?

Due to the RSTUF Worker changes in the database models, the feature `rstuf import-targets` no long works.

There was breaking compatibility, and this functionality no longer works.

* The `RSTUFTargets` table was changed to `RSTUFTargetFiles`
* The `RSTUFTargetFiles` has a new field `targets_roles` which is the `FK` of `RSTUFTargetRoles`

The problem is that the `import-targets` requires now to give the `RSTUFTargetRoles` when inserting the row

### What steps did you take?

Try to use `import-target` will fail

### What behavior did you expect?

_No response_

### Relevant log output

_No response_

### Code of Conduct

kairoaraujo:
This commit fixes the bug on `rstuf admin import-targets`, which wasn't working with RSTUF Worker/API beta version.

The problem is due to the changes in database models for RSTUF Workers.

It also simplifies the way to retrieve the succinct roles.

Closes #297

### DIFF
--- a/repository_service_tuf/cli/admin/import_targets.py
+++ b/repository_service_tuf/cli/admin/import_targets.py
@@ -1,20 +1,21 @@
 import json
 import os
 from datetime import datetime
+from math import log
 from typing import Any, Dict, List
 
-from tuf.api.metadata import Delegations, SuccinctRoles, Targets
+from tuf.api.metadata import SuccinctRoles
 
 from repository_service_tuf.cli import click, console
 from repository_service_tuf.cli.admin import admin
 from repository_service_tuf.helpers.api_client import (
+    URL,
     Methods,
     bootstrap_status,
     publish_targets,
     request_server,
     task_status,
 )
-from repository_service_tuf.helpers.tuf import Metadata
 
 
 def _check_csv_files(csv_files: List[str]):
@@ -30,23 +31,33 @@ def _check_csv_files(csv_files: List[str]):
 
 
 def _parse_csv_data(
-    csv_file: str, succinct_roles: SuccinctRoles
+    db: Any,
+    rstuf_target_roles: Any,
+    succinct_roles: SuccinctRoles,
+    csv_file: str,
 ) -> List[Dict[str, Any]]:
     rstuf_db_data: List[Dict[str, Any]] = []
     with open(csv_file, "r") as f:
         for line in f:
+            path = line.split(";")[0]
+            length = int(line.split(";")[1])
+            hash_algorithm = line.split(";")[2]
+            hash_digest = line.split(";")[3]
             rstuf_db_data.append(
                 {
-                    "path": line.split(";")[0],
+                    "path": path,
                     "info": {
-                        "length": int(line.split(";")[1]),
-                        "hashes": {line.split(";")[2]: line.split(";")[3]},
+                        "length": length,
+                        "hashes": {hash_algorithm: hash_digest},
                     },
-                    "rolename": succinct_roles.get_role_for_target(
-                        line.split(";")[0]
-                    ),
                     "published": False,
                     "action": "ADD",
+                    "targets_role": db.execute(
+                        rstuf_target_roles.select().where(
+                            rstuf_target_roles.c.rolename
+                            == succinct_roles.get_role_for_target(path)
+                        )
+                    ).one()[0],
                     "last_update": datetime.now(),
                 }
             )
@@ -56,7 +67,8 @@ def _parse_csv_data(
 
 def _import_csv_to_rstuf(
     db_client: Any,
-    rstuf_table: Any,
+    rstuf_target_files: Any,
+    rstuf_target_roles: Any,
     csv_files: List[str],
     succinct_roles: SuccinctRoles,
 ) -> None:
@@ -65,10 +77,12 @@ def _import_csv_to_rstuf(
 
     for csv_file in csv_files:
         console.print(f"Import status: Loading data from {csv_file}")
-        rstuf_db_data = _parse_csv_data(csv_file, succinct_roles)
+        rstuf_db_data = _parse_csv_data(
+            db_client, rstuf_target_roles, succinct_roles, csv_file
+        )
         console.print(f"Import status: Importing {csv_file} data")
         try:
-            db_client.execute(rstuf_table.insert(), rstuf_db_data)
+            db_client.execute(rstuf_target_files.insert(), rstuf_db_data)
         except IntegrityError:
             raise click.ClickException(
                 "Import status: ABORTED due duplicated targets. "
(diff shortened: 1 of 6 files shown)


repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -65,10 +77,12 @@ def _import_csv_to_rstuf(
...
     for csv_file in csv_files:
         console.print(f"Import status: Loading data from {csv_file}")
-        rstuf_db_data = _parse_csv_data(csv_file, succinct_roles)
+        rstuf_db_data = _parse_csv_data(
+            db_client, rstuf_target_roles, succinct_roles, csv_file
+        )
         console.print(f"Import status: Importing {csv_file} data")
         try:
-            db_client.execute(rstuf_table.insert(), rstuf_db_data)
+            db_client.execute(rstuf_target_files.insert(), rstuf_db_data)
```
MVrachev, line 85: So, this will add a new record to the `rstuf_target_files` table and the db client will put the `rstuf_db_data` on its place?

MVrachev (commented):
I will look at the tests in the second review.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -78,27 +92,35 @@ def _import_csv_to_rstuf(
...
+            f"Failed to retrieve RSTUF config {response.text}"
+        )
 
-    json_data = json.loads(response.text)
-    targets: Metadata[Targets] = Metadata.from_dict(json_data)
-    if targets.signed.delegations is None:
-        raise click.ClickException("Failed to get Targets Delegations")
+    try:
+        data = response.json()["data"]
+        num_bins = data["number_of_delegated_bins"]
```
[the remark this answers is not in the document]
kairoaraujo: It is a bug in the API documentation.
Screenshot 2023-06-28 at 15 39 04

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: The `rstuf_db_data` is the data to be added to the DB, but by default, it is a Python dictionary where the key is the column name.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -56,7 +67,8 @@ def _parse_csv_data(
 
 def _import_csv_to_rstuf(
     db_client: Any,
```
MVrachev, line 69: Can you improve the annotation to be more precise?
From what I see `db_client` is a `sqlaclehmy` object called `Connection`.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -30,23 +31,33 @@ def _check_csv_files(csv_files: List[str]):
 
 
 def _parse_csv_data(
-    csv_file: str, succinct_roles: SuccinctRoles
+    db: Any,
```
MVrachev, line 34: Can you improve the annotation to be more precise?
From what I see `db` is a `sqlaclehmy` object called `Connection`.

repository_service_tuf/cli/admin/import_targets.py
MVrachev: I think I got what this line means.
We ask the `db` connection to execute a task or to `insert` what is given as a second argument or `rstuf_db_data`.
Am I correct?

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -78,29 +92,37 @@ def _import_csv_to_rstuf(
...
+    except (json.JSONDecodeError, KeyError):
+        raise click.ClickException(
+            "Failed to parse 'data', 'number_of_delegated_bins' from config "
+            f"{response.text}"
+        )
+    bit_length = int(log(num_bins, 2))
 
-    if targets_delegations.succinct_roles is None:
-        raise click.ClickException("Failed to get Targets succinct roles")
+    # the 'keyids' and the 'threshold' are irrelevant once we need the names
```
MVrachev, line 113: I don't understand what you mean by this comment.
Maybe something like this: 
```suggestion
    # 'threshold' is always 1 as we use a single online key 'keyids' will be
    # populated  when the online key is added
```

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -140,80 +176,81 @@ def test__import_csv_to_rstuf_duplicate_targets(self):
...
+    def test__get_succinct_roles(self):
+        fake_response = pretend.stub(
+            status_code=200,
+            json=pretend.call_recorder(
+                lambda: {"data": {"number_of_delegated_bins": 16}}
+            ),
         )
-        monkeypatch.setattr(
-            import_targets.json, "loads", lambda *a: "json_data"
+        import_targets.request_server = pretend.call_recorder(
```
MVrachev, line 201: Maybe it's not a good idea to pretend directly the `import_targets` and instead use `monkepatch`?
I recently had problems with functions still being `pretend` after the test execution.

MVrachev (commented):
Did you test this fix with an RSTUF API and RSTUF Worker deployment?

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: Yes, that is correct.

kairoaraujo:
> Did you test this fix with an RSTUF API and RSTUF Worker deployment?

Yes, I tested in a running evironment

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: Even if we decide to use multiple online keys or using offline keys for the targets, it is still irrelevant as we don't do operations to the metadata, only the database.

tests/unit/cli/admin/test_import_targets.py
kairoaraujo: It is required only if `import_target` is used in another scope.
In the case of the CLI, used by multiple command lines namespaces, like your case.

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: Oh, I notice something.

If I do it, I will need to import Connection and Table outside the CLI command.
The SQLAlchemy will become a required [instead of an optional dependency](https://repository-service-tuf.readthedocs.io/en/v1.0.0a1-draft/guide/repository-service-tuf-cli/index.html#import-targets-import-targets).

repository_service_tuf/cli/admin/import_targets.py
kairoaraujo: See[ my comment above](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/304#discussion_r1246452172).

tests/unit/cli/admin/test_import_targets.py
```diff
@@ -98,25 +123,36 @@ def test__import_csv_to_rstuf(self):
...
+                fake_rstuf_files.insert(), [{"k1": "v1", "k2": "v2"}]
             ),
             pretend.call(
-                fake_rstuf_table.insert(), [{"k1": "v1", "k2": "v2"}]
+                fake_rstuf_files.insert(), [{"k1": "v1", "k2": "v2"}]
             ),
         ]
-        assert fake_rstuf_table.insert.calls == [
+        assert fake_rstuf_files.insert.calls == [
             pretend.call(),
```
MVrachev, line 156: Any idea why `fake_rstuf_files.insert.calls` calls are three when there are two `csv` files?

MVrachev (approved):
I have a single question, otherwise it's great.

tests/unit/cli/admin/test_import_targets.py
kairoaraujo: I am fixing it.
It is because I was also calling it on lines 149 and 152.

MVrachev:
I think the failing of the CI is the same as I saw in my pr #284.
I fixed it in the commit https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/284/commits/c594160f0e623e86a4cae94f6c69b83677d8a3e2, but I have some questions regarding this fix as I shared in my following comment: https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/284#issuecomment-1620240870

PR #322, opened 2023-07-17
Add shell completion automatically to `rstuf`

KAUTH:
Test the CLI: `pip install -e .`

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,7 @@ rich-click = "*"
 securesystemslib = {extras = ["crypto"], version = "*"}
 click = "*"
 rich = "*"
+auto-click-auto = "*"
 PyNaCl = "==1.5.0"
 requests = "*"
 tuf = "==3.0.0"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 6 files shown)


kairoaraujo:
Isn't required in that case to have it in the `pyproject.toml`?

KAUTH:
> Isn't required in that case to have it in the pyproject.toml?

Yes, I forgot to add it there, thanks!

kairoaraujo:
Not sure why, it doesn't work on my terminal (ksh) 😞

KAUTH:
> I don't know why, but after I install rstuf cli with `pip install -e .` still the terminal doesn't suggest autocomplete options for rstuf.
I am using `zsh`.

- Can you run on you shell: `echo $SHELL`? If it's not `zsh` it's probably because `bash` is set to the default shell.
- In any case, I think there's an easy fix for this. I pushed the change

MVrachev:
> * Can you run on your shell: `echo $SHELL`? If it's not `zsh` it's probably because `bash` is set to the default shell.

```
(.env) ➜  cli [17:47]:  echo $SHELL
/bin/zsh
```

> * In any case, I think there's an easy fix for this. I pushed the change, can you retry?

I tried it again, but sadly it doesn't work.
When you have time we can debug this together.

kairoaraujo (approved):

PR #329, opened 2023-07-31
Updated __init__.py for Feature: Custom timeout for Ceremony Bootstrap

primuzer:


kairoaraujo (changes requested):
Hi @primuzer,

Thank you for your contribution to the RSTUF project 🎉 

The issue, my fault, is not clear enough, and I see that

We need to add the parameter `-t/--timeout`, but it needs to be part of the `rstuf admin ceremony` part, which is on the ceremony module.
https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/f4b70476096b239eab06bfefd89894c41c2f0007/repository_service_tuf/cli/admin/ceremony.py#L638

Once we have the parameter `-t/--timeout` we will need also to add it to the payload, if passed by the user.

https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/f4b70476096b239eab06bfefd89894c41c2f0007/repository_service_tuf/cli/admin/ceremony.py#L691
```python 
if timeout:
  json_payload["timeout"] = timeout
```

It will also require implement the UT in the https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/f4b70476096b239eab06bfefd89894c41c2f0007/tests/unit/cli/admin/test_ceremony.py#L336

Also, please remember to [sign to accept DCO as mentioned here](https://repository-service-tuf.readthedocs.io/en/stable/devel/contributing.html#dco).

PR #340, opened 2023-08-09
Distributed Asynchronous Signing - Bootstrap

kairoaraujo:
This PR has three main commits.

### Refactoring: Rename `RootInfo` to `MetadataInfo`

It enables to reuse of the `MetadataInfo` class from Metadata Update to the Metadata Sign feature.
It renames some functions from this class too.
It doesn't rename some tests because the 'root' is a use case.

### Implement the `rstuf admin metadata sign`

This command allows the user to sign.

### Implement the `rstuf admin ceremony` 

The private key is required by whoever is starting the ceremony, to at least have one valid signature.
For the other root keys, there is a possibility to choose between the private or public key. The public key gives the possibility to sign later the metadata using the `rstuf admin metadata sign`

Closes #294

PR #348, opened 2023-08-15
Add `--token` option for `rstuf` commands

KAUTH:
- This option allows the user to pass an RSTUF API authentication token,
    besides the "admin" one.
    If the `--auth` option is provided this token is not used for
    authentication.
- `rstuf admin metadata` and `rstuf admin ceremony` have been adjusted to accept the `--token` option.
- **Please also test that the `ceremony` and `metadata` commands work as expected with the `--token` option.**

### DIFF
--- a/repository_service_tuf/helpers/api_client.py
+++ b/repository_service_tuf/helpers/api_client.py
@@ -23,6 +23,7 @@ class URL(Enum):
     metadata = "api/v1/metadata/"
     task = "api/v1/task/?task_id="
     publish_targets = "api/v1/targets/publish/"
+    artifacts = "api/v1/artifacts/"
 
 
 class Methods(Enum):
@@ -64,11 +65,11 @@ def request_server(
     return response
 
 
-def is_logged(settings: LazySettings):
-    if settings.get("AUTH") is False:
+def is_logged(settings: LazySettings, token: Optional[str] = None):
+    if settings.get("AUTH") is False and token is None:
         return None
 
-    token = settings.get("TOKEN")
+    token = settings.get("TOKEN") or token
     server = settings.get("SERVER")
     headers = {"Authorization": f"Bearer {token}"}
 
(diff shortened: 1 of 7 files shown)


repository_service_tuf/helpers/api_client.py
```diff
@@ -23,6 +23,7 @@ class URL(Enum):
     metadata = "api/v1/metadata/"
     task = "api/v1/task/?task_id="
     publish_targets = "api/v1/targets/publish/"
+    artifacts = "api/v1/artifacts/"
```
kairoaraujo, line 26: nit: I think it is part of you `rstuf artifacts add` PR

repository_service_tuf/helpers/api_client.py
```diff
@@ -88,24 +89,38 @@ def is_logged(settings: LazySettings):
...
+) -> Dict[str, str]:
+
+    built_in_auth: bool = settings.get("AUTH")
+
+    if built_in_auth is False and token is None:
         return {}
 
     server = settings.get("SERVER")
-    token = settings.get("TOKEN")
+    token = settings.get("TOKEN") or token
```
kairoaraujo, line 102: Verifying the opposite, the `token` and the `settings` is better.get("TOKEN").
There is a situation where the user has `.rstuf.ini` with the TOKEN, but he didn't call `--auth`; in that scenario, it will load the token from the setting file.
```suggestion
    token = token or settings.get("TOKEN")
```

repository_service_tuf/helpers/api_client.py
```diff
@@ -88,24 +89,38 @@ def is_logged(settings: LazySettings):
...
-        token_access_check = is_logged(settings)
+        additional_error_message: str = ""
+
+        token_access_check = is_logged(settings, token)
         if token_access_check.state is False:
+            if built_in_auth is True:
+                additional_error_message = (
+                    "\n\nTry re-login: 'rstuf --auth admin login'"
+                )
+
```
kairoaraujo, line 112: If the status code is 401 or 403, the data is none. I see you handle it in case of built-in auth but not token.
Could you include an else here? 
If  `token_access_check.state` is `False`, and the `built_in_auth` is `False`, the user will see only `None`.
```suggestion
            if built_in_auth is True:
                additional_error_message = (
                    "\n\nTry re-login: 'rstuf --auth admin login'"
                )
            else:
                 additional_error_message = "Token is invalid"

```

kairoaraujo (changes requested):

kairoaraujo:
I see we have more `get_headers` where we will also need to pass the `token` in the `api_client`
Example:
https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/383b020e5eb9b8a1b002ea20093e059647498cbe/repository_service_tuf/helpers/api_client.py#L143

repository_service_tuf/helpers/api_client.py
KAUTH: Good catch! Hm, but we are still missing a case, from #39 we have that: 
> ... if the user uses `rstuf --auth` it ignores the token parameter.

this means that the priority is --`auth` > `--token`, if both are there, which makes sense. But with the suggested snippet `token` will be chosen. I have changed the code to take care of this case as well.

kairoaraujo:
Hi @KAUTH, I'm proposing [here a simplified way to support ](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/340/commits/185dc90be290d837f9f3432631c06660532b2e7a)`-t/--token` which makes it compatible with all commands/subcommands.

KAUTH:
Closing the PR in favor of https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/340/commits/185dc90be290d837f9f3432631c06660532b2e7a.

PR #369, opened 2023-09-07
Adds  `-t/--token` parameter

kairoaraujo:
# Description

The paramenter `-t/--token` can be user in combination with admin commands.

It covers the cases where the administrator gives an RSTUF command user a token to allow to run specific administrative sub-commands.

Example:

The admin starts a bootstrap authenticated as admin (`--auth`) The bootstrap requires other key holds to sign the root metadata.

Admin generates tokens allowing key holders to send requests to the signing process endpoints.

The key holders can run `rstuf --token <token> admin metadata sign` Add support to `-t/--token`

Fixes #(issue)

# Types of changes
- [X] New feature (non-breaking change which adds functionality)

# Additional requirements

# Code of Conduct


repository_service_tuf/helpers/api_client.py
```diff
@@ -90,38 +92,32 @@ def is_logged(settings: LazySettings):
...
-                    f"Unexpected error: {response.text}"
-                )
-    else:
+
+    elif settings.get("AUTH") is True:
         raise click.ClickException(
             "Login first. Run 'rstuf --auth admin login'"
         )
 
+    else:
```
KAUTH, line 118: At this point it means that we have:
- no `--auth` given or
- no `--token` given
- or both
Since the function is `get_headers` it would make more sense to me to return an exception with a message to give either --auth or --token (e.g., "Provide the `--auth` or `--token` option"), because we are trying to actually get the headers.

KAUTH (changes requested):
Very straightforward! I only have one comment to discuss.

KAUTH (approved):
LGTM!

Issue #355, opened 2023-08-22
Task: add information about pending signatures during a DAS Bootstrap ceremony


MVrachev:
### What is the task about?

Add information to the user about pending signatures during a DAS Bootstrap ceremony.
I imagine this will be at the end of the bootstrap ceremony so that the user understands that:
1. the root metadata is not yet considered trustworthy
2. X amount of signatures are needed to finish the bootstrap process

This is important as there are multiple questions and steps in the bootstrap ceremony and it's possible to forget what was the threshold and how many keys were actually loaded and the user can be left with the impression he has finished with the bootstrap ceremony.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ivanayov:
Can I work on this?

PR #405, opened 2023-09-29
Check if root keys are not less than a threshold

ivanayov:
This change adds a check during the bootstrap ceremony that the root keys are not less than the root threshold and outputs a warning otherwise

Follow-up of #355 and #377

# Types of changes
- [n/a ] New feature (non-breaking change which adds functionality)

# Additional requirements
- [n/a] Docs have been added for the bug fix or new feature

# Code of Conduct


repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -638,6 +638,15 @@ def _run_ceremony_steps(save: bool) -> Dict[str, Any]:
             f" {pending_signatures} pending signature(s) left.",
         )
 
+    required_num_of_keys = (
+        setup.threshold[Roles.ROOT] - setup.number_of_keys[Roles.ROOT]
```
MVrachev, line 642: It's important to be clear with the terminology here.
For DAS we expect that the number of **`LOADED`** keys will be less than the threshold, but the number of keys that we have **`SETUP`** for root should be equal to or more than the threshold.
The number of keys we have **`SETUP`** is equal to the number of  **`LOADED`** keys + the number of keys for which we have provided public key information.

Here we want to print the difference between how many keys we have **`SETUP`** minus the number of expected root keys in total.
I think you shouldn't use here `setup.threshold[Roles.ROOT]`, but instead `len(metadata[Roles.ROOT.value].signed.keys)`.
The `metadata[Roles.ROOT.value].signed.keys` contains all **`LOADED`**  keys + those for which we have given public key value.

Am I correct with my understanding here @kairoaraujo?

MVrachev: Another mistake I see: it should be `setup.number_of_keys[Roles.ROOT] - len(metadata[Roles.ROOT.value].signed.keys)` otherwise the result will be 0 or negative.

kairoaraujo: > Am I correct with my understanding here @kairoaraujo?

Yes, It is correct. I'm adding some simplification.

> It's important to be clear with the terminology here. For DAS we expect that the number of **`LOADED`** keys will be less than the threshold, but the number of keys that we have **`SETUP`** for root should be equal to or more than the threshold.

IMO, we don't need to care if it is DAS or Signed process, but check what you suggested below.

> The number of keys we have **`SETUP`** is equal to the number of **`LOADED`** keys + the number of keys for which we have provided public key information.
> 
> Here we want to print the difference between how many keys we have **`SETUP`** minus the number of expected root keys in total. I think you shouldn't use here `setup.threshold[Roles.ROOT]`, but instead `len(metadata[Roles.ROOT.value].signed.keys)`. The `metadata[Roles.ROOT.value].signed.keys` contains all **`LOADED`** keys + those for which we have given public key value.
>

ivanayov: This use-case, i.e. `(A)setup.number_of_keys[Roles.ROOT] - (B)len(metadata[Roles.ROOT.value].signed.keys)` will never be >0 as `signed.keys` includes all the keys, not only the ones used to sign root. There will always be at least one online key. 
* If (A) == 1 (instead of the default 2), (A) - (B) will be <0, as (B) will be at least 1 root key (ceremony would fail prior otherwise) + at least 1 online key. 
* If (A) == 2, (B) would be again >= 2 (at least 1 root key + at least 1 online key)
* If (A) > 2, ceremony would fail prior without having configured a metadata, that has >2 root keys - one cannot give a setup number of root keys without passing them as ceremony input afterwards, so this will result to a metadata that has (A) number of keys in and we go to the previous 2 use-cases - we'd just increase the numbers.

If we use only the loaded keys used to sign root for (B) instead of `metadata[Roles.ROOT.value].signed.keys`, (A)-(B) will always be 0, as during ceremony one would pass as many root keys as (A) to generate the root metadata. Otherwise, ceremony would fail prior this check.

This makes a so defined `required_num_of_keys` useless. 
This is why it's now changed to compare threshold to the loaded keys used to sign root, instead of `setup.number_of_keys[Roles.ROOT]`, only for functional clarity, even though they would be practically equal.

ivanayov: In this context `required_num_of_keys` will always be equal to `pending_signatures` out of the DAS context.

Moved this check prior the check for `pending_signatures`, as the latter will always be >=`required_num_of_keys` in the DAS context.

ivanayov: We had a live chat with @MVrachev and both agreed on this approach.

@kairoaraujo Could you please share your point on that? Thanks

MVrachev: > If (A) > 2, ceremony would fail prior for not having configured a metadata, that has >2 root keys - one cannot give a setup number of root keys without passing them as ceremony input afterwards, so this will result to a metadata that has (A) number of keys in and we go to the previous 2 use-cases - we'd just increase the numbers.

About this case: we make it mandatory to input the same amount of keys as you have entered as `number_of_keys`.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -625,9 +625,19 @@ def _run_ceremony_steps(save: bool) -> Dict[str, Any]:
...
+    root_role = root_metadata.signed.roles[Roles.ROOT.value]
+    required_num_of_keys = root_role.threshold - len(root_role.keyids)
+    if required_num_of_keys > 0:
+        console.print(
+            "Not enough keys set for root,"
+            f" {required_num_of_keys} more key(s) left to reach threshold.",
+        )
+
     # Inform user for pending signatures
     pending_signatures: int = setup.threshold[Roles.ROOT] - len(
```
MVrachev, line 639: In order to make it complete
```suggestion
    pending_signatures: int = root_role.threshold - len(
```

ivanayov:
Thanks @MVrachev and @kairoaraujo :) I've addressed latest comments and updated pre-commit, as the older version was causing version check failures.

MVrachev (approved):

Issue #332, opened 2023-08-01
Feature:  CLI artifact add command `rstuf artifact download` 


kairoaraujo:
### What is the task about?

This task is about creating the download feature for `rstuf artifact`

It should use the python-tuf ngclient. The [python-tuf client](https://github.com/theupdateframework/python-tuf/tree/develop/examples/client) example is a good reference implementation

## Download Artifact (Feature)

Note: The download feature ignores `rstuf --auth`, it uses the public metadata and the repository information. It doesn't use the RSTUF API.

### Settings
The `rstuf` command line has the `rstuf.ini` and it must be used for the download feature.
https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/9df308177620245cdbbbb401a1217a3e2d5df5bb/repository_service_tuf/cli/__init__.py#L63-L67

#### Trusted Root (`root`)

The user can configure the on the `rstuf.ini` the Metadata URL or use a parameter `-r/--root`
This is the path of the initial trusted root TUF Metadata

#### Metadata URL (`metadata_url`)

The user can configure the on the `rstuf.ini` the Metadata URL or use a parameter `-m/--metadata_url`
This is the URL where the python-tuf ngclient will check the metadata

#### Artifact Base URL (`artifact_url`)

The user can configure the on the `rstuf.ini` the Artifact Base URL or use a parameter `-a/--artifacts_url`
This URL is used to fetch the artifacts by the python-tuf ngclient
Example:

`rstuf artifact download project/file1.tar.gz`
The ngclient will use `http://downloads.example.com` as the artifact URL (target base URL)  and the path `project/file1.tar.gz`, fetching from `http://downloads.example.com/project/file1.tar.gz` when the target is valid.

####  Hash Prefix (`hash_prefix`)

The user can configure the on the `rstuf.ini` the `hash_prefix` True or False or use a parameter `--hash-prefix`
If the user uses --hash-prefix it becomes `True`

The default (if not configured also) is `False`

It is used to add the hash or not.
The above example (`rstuf artifact download --hash-prefix project/file1.tar.gz`) will become `http://downloads.example.com/project/<hash>-file1.tar.gz`
An example of this implementation is [available here](https://github.com/theupdateframework/python-tuf/blob/12d0c3cd1e168d07fcbccca8c774e658b5562a97/examples/client/client#L76)

### Parent feature

- https://github.com/repository-service-tuf/repository-service-tuf-cli/issues/334

[...]

PR #408, opened 2023-10-03
Remove authnz

### LINKED ISSUE #406: Task: Remove authn/z and token feature from `rstuf` CLI
### What is the task about?

This task is about removing all the features for authentication and authorization from the `rstuf` command line interface.

### Parent feature

- https://github.com/repository-service-tuf/repository-service-tuf/issues/491

### References

_No response_

### Code of Conduct

kairoaraujo:
# Description

Removes Command Line Interface for the RSTUF API built-in authentication
and authorization (authn/z).

The authn/z will be removed from the RSTUF API, it is related to
https://github.com/repository-service-tuf/repository-service-tuf/issues/491

This commit removes the implementation for `--auth` and `--token` in the
code base and its tests.

Adds consistency to the `--api-url` parameter.

Remove auth, login, token from documentation

Fixes #406 

# Types of changes

# Additional requirements

# Code of Conduct


repository_service_tuf/cli/admin/metadata.py
```diff
@@ -628,17 +626,16 @@ def update(
...
+    settings: Any, api_server: Optional[str]
 ) -> Dict[str, Any]:
-    if settings.AUTH is False and api_url is None:
-        api_url = prompt.Prompt.ask("\n[cyan]API[/] URL address")
-        settings.SERVER = api_url
-    elif settings.AUTH is False and api_url is not None:
-        settings.SERVER = api_url
+    if api_server is None:
+        api_server = prompt.Prompt.ask("\n[cyan]API[/] URL address")
+        settings.SERVER = api_server
```
MVrachev, line 633: If we have used the `--config` option and `--api-server` and there is a server URL in the config file this means that the `--api-server` option will have a higher priority right?

MVrachev (commented):
I noticed a couple of other places that needs to be removed:
- https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/repository_service_tuf/helpers/api_client.py#L20
- the mention of `token` here: https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/README.rst?plain=1#L16


I am a little confused about the `config` file. I see we store the server address where we send our payloads, but this file is generated after the `login` command which I think we should remove: 
- https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/docs/source/devel/design.rst?plain=1#L20
- https://github.com/kairoaraujo/repository-service-tuf-cli/blob/908dd5885744b53805fcea245a26218fc2c897dc/repository_service_tuf/helpers/api_client.py#L36

My confusion is whether we need it at all?
I checked how we use it and what we do is we load the config file into `Dynaconf` settings object and from then on it seems we only assign values to the `settings.SERVER` attribute.
It seems to me that we can replace the `settings.SERVER` usage with a simple local `api_url` variable instead.
What do you think?

kairoaraujo:
Thank you for the review @MVrachev 

> I noticed a couple of other places that needs to be removed:
> 
> * https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/repository_service_tuf/helpers/api_client.py#L20
> * the mention of `token` here: https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/README.rst?plain=1#L16

Fixed

> I am a little confused about the `config` file. I see we store the server address where we send our payloads, but this file is generated after the `login` command which I think we should remove:
> 
> * https://github.com/kairoaraujo/repository-service-tuf-cli/blob/1097d527813ab563d2be8b13fc012fad69eb91c2/docs/source/devel/design.rst?plain=1#L20

I changed that to reads. We will change it even more when we have the #332 

> * https://github.com/kairoaraujo/repository-service-tuf-cli/blob/908dd5885744b53805fcea245a26218fc2c897dc/repository_service_tuf/helpers/api_client.py#L36

> My confusion is whether we need it at all? I checked how we use it and what we do is we load the config file into `Dynaconf` settings object and from then on it seems we only assign values to the `settings.SERVER` attribute. It seems to me that we can replace the `settings.SERVER` usage with a simple local `api_url` variable instead. What do you think?

It allows the user to create a file with `SERVER` so he doesn't need to use the parameter every time.

Having priority (`--api-server`) is good because you have a configured server (`config`), but you need to do a one-time operation to another API server, but maybe add it as a good first issue?

docs/source/devel/design.rst
```diff
@@ -16,11 +16,11 @@ Container level
 The ``repository-service-tuf``, in the container perspective, is a command line tool that
 interacts to the ``repository-service-tuf-api``.
 
-``repository-service-tuf`` writes a settings configuration in the file
-``$HOME/.rstuf.ini`` with ``login`` subcommand.
+``repository-service-tuf`` reads settings configuration in the file
```
MVrachev, line 19:
```suggestion
``repository-service-tuf`` reads the settings configuration from path
```

repository_service_tuf/helpers/cli.py
```diff
@@ -75,10 +74,9 @@ def create_artifact_payload_from_filepath(
 ) -> Dict[str, Any]:
     """
     Create the payload for the API request of `POST api/v1/artifacts/`.
-
     The blake2b-256 cryptographic hash function is used to hash the file.
 
-    :param filepath: path to the artifact to be added to the metadata.
+    :param filepath: The file path to calculate the hash.
```
MVrachev, line 79: For future: please make such documentation changes in separate commits and document that this is required from `make docs`. I am asking this as those changes are not related to the main changes in any way.
I had the same problem before.

repository_service_tuf/cli/admin/metadata.py
```diff
@@ -709,18 +706,16 @@ def _sign_metadata(role_info: MetadataInfo, rstuf_key: RSTUFKey) -> Signature:
 def _delete_signing_process(
     role: str, settings: Any, api_url: Optional[str]
 ) -> None:
-    if settings.AUTH is False and api_url is None:
+    if api_url is None:
         api_url = prompt.Prompt.ask("\n[cyan]API[/] URL address")
```
MVrachev, line 710: Why do you request a user to enter `API URL` if he has already setup `settings.SERVER`? 
Also, `_delete_signing_process` is called from the `sign` command where `api_url` is optional, but how can it be optional where we check if it's `None` and if it is we require value?

repository_service_tuf/cli/admin/metadata.py
```diff
@@ -529,21 +527,21 @@ def update(
...
     settings = context.obj["settings"]
     if upload and not run_ceremony:
         # Sever authentication or setup
-        if settings.AUTH and not upload_server:
+        if api_server is None:
             raise click.ClickException(
-                "Requires '--upload-server' when using '--auth'. "
-                "Example: --upload-server https://rstuf-api.example.com"
+                "Requires '--api-server' when using '--upload/-u'. "
+                "Example: --api-server https://api.rstuf.example.com"
```
MVrachev, line 541: Why throw an error when the user could have setup `settings.SERVER` to be used meaning `api-url` is not required?
Also, `--api-server` is documented as `required=False`, so it should be optional.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -724,13 +723,13 @@ def ceremony(
 
     # option bootstrap: checks if the server accepts it beforehand
     if bootstrap:
-        if settings.AUTH is False and upload_server is None:
+        if api_server is None:
```
MVrachev, line 726: Same question as for the `update` command.

repository_service_tuf/cli/admin/import_targets.py
```diff
@@ -120,7 +120,7 @@ def _get_succinct_roles(api_url: str) -> SuccinctRoles:
 
 @admin.command()  # type: ignore
 @click.option(
-    "--api-url",
+    "--api-server",
     required=True,
```
MVrachev, line 124: Why `--api-server` is required if the user can setup `settings.SERVER` to be used?

MVrachev (commented):
I need some clarifications between `settings.SERVER` and `api-url`.

On another note: why did you add `rstuf.yaml` file?

repository_service_tuf/cli/admin/metadata.py
```diff
@@ -709,18 +706,16 @@ def _sign_metadata(role_info: MetadataInfo, rstuf_key: RSTUFKey) -> Signature:
 def _delete_signing_process(
     role: str, settings: Any, api_url: Optional[str]
```
MVrachev, line 707: I just noticed `role` can be removed here as it's not used anywhere.

kairoaraujo:
> I need some clarifications between `settings.SERVER` and `api-url`.

The paramenter will be asked to the user when there is no  `rstuf.ini` wih `SERVER`.
But, on my machine, for example, I can have it and I don't need use it everytime.

It will become more explicity when we improve the UX and start having features such as `rstuf artifacts` commands.

> On another note: why did you add `rstuf.yaml` file?

This file was added by mistake. I sent a fixup for it.

docs/source/devel/design.rst
MVrachev: Nit: the above comment.

repository_service_tuf/cli/admin/import_targets.py
MVrachev: Will love to hear your thoughts here.

MVrachev (commented):
A nit and a question remain.

kairoaraujo:
Rebased

MVrachev (approved):

kairoaraujo:
https://github.com/kairoaraujo/repository-service-tuf-cli/actions/runs/6479813035

Functional Tests using the Umbrella fix https://github.com/repository-service-tuf/repository-service-tuf/pull/521

MVrachev (approved):

PR #416, opened 2023-10-10
Add `rstuf artifact download` command

### LINKED ISSUE #332: Feature:  CLI artifact add command `rstuf artifact download`

### LINKED ISSUE #433: Task: Implement `rstuf artifact download` command
### What is the task about?

The command should be with the following layout:
```
❯ rstuf -c rstuf.yaml artifact download --help
                                                                                                  
 Usage: rstuf artifact download [OPTIONS] ARTIFACT                                                
                                                                                                  
 Downloads artifacts to the TUF metadata using python-tuf ``ngclient``.                           
                                                                                                  
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────╮
│ --root      -t  TEXT  A metadata URL to the initial trusted root.                              │
│ --metadata-url      -   TEXT  A URL to download TUF metadata from.                             │
│ --artifacts-url     -a  TEXT  An artifacts base URL to fetch from.                             │
│ --hash-prefix       -h        A hash prefix.                                                   │
│ --directory-prefix  -P  TEXT  A prefix for the download dir.                                   │
│ --help                        Show this message and exit.                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Parent feature

#332

### References

_No response_

### Code of Conduct

ivanayov:
This change adds an `rstuf artifact download` command with the following usage:

```
Downloads artifacts to the TUF metadata using python-tuf
``ngclient``.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
│ --root              -r  TEXT  A metadata URL to the initial trusted root or a local file.        │
│ --metadata-url      -m  TEXT  TUF Metadata repository URL.                                       │
│ --artifacts-url     -a  TEXT  An artifacts base URL to fetch from.                               │
│ --hash-prefix       -p        A hash prefix.                                                     │
│ --directory-prefix  -P  TEXT  A prefix for the download dir.                                     │
│ --help              -h        Show this message and exit.                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Fixes #433

# Types of changes

# Additional requirements

# Code of Conduct


repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+    "-t",
+    "--trusted-root",
+    "trusted_root",
+    help="A metadata URL to the initial trusted root.",
+    type=str,
+    required=None,
+    default=None,
+)
+@click.option(
+    "-",
```
ivanayov, line 52: Noticed that 😄

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+        console.print(
+            "Trusted local root not found. Using 'tofu' to "
+            "Trust-On-First-Use or copy trusted root metadata to "
+            f"{metadata_dir}/root.json"
+        )
+        ok = init_tofu(metadata_url, trusted_root)
+        if not ok:
+            console.print(
+                f"Trusted local root not found in {metadata_url}"
+                "`tofu` was not successful"
```
ivanayov, line 225: Should we use 'tofu' here when it's not a subcommand or rather explain the approach?

repository_service_tuf/cli/artifact/repository.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: 2023 VMware Inc
```
ivanayov, line 1: Preferred to not mess a repository logic into the download client. 

Do you think it fits together with the upcoming `rstuf artifact repository` command?
It's again a client, so having the hash bin logic here may be confusing as well.

ivanayov:
@kairoaraujo What do you think about this approach?

It makes the config optional and allows multi-repo logic if it's used,
but overwrites it if any of the config options are given as args.

Allows download with tofu only without a config file, but with args instead (as it makes no sense to have a config before having initial trusted root)
and fails if a config is used, but trusted root is not set and available.

May be some of the flags should be required? Now one can run `rstuf artifact download -n`, for example, without specifying metadata and artifacts urls, so at least they should be required, but IMO all others can be optional.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+            artifacts_url = rstuf_config["repositories"][repository][
+                "artifacts_url"
+            ]
+        if not metadata_url:
+            metadata_url = rstuf_config["repositories"][repository][
+                "metadata_url"
+            ]
+        if not hash_prefix:
+            hash_prefix = rstuf_config["repositories"][repository][
+                "hash_prefix"
```
ivanayov, line 135: A note to myself to handle missing configuration =)

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+)
+@click.option(
+    "-n",
+    "--arifact-name",
+    "artifact_name",
+    help="An artifact name to get.",
+    type=bool,
+    required=None,
+    default=None,
+)
```
kairoaraujo, line 77: I think we don't want it, too many options and it should be an argument.

So the user can simply run

```
rstuf -c rstuf.yaml artifact download my-app.1.0.4.tar.gz
```

```shell
❯ rstuf -c rstuf.yaml artifact download --help
                                                                                                  
 Usage: rstuf artifact download [OPTIONS] ARTIFACT                                                
                                                                                                  
 Downloads artifacts to the TUF metadata using python-tuf ``ngclient``.                           
                                                                                                  
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────╮
│ --repository        -r  TEXT  A repository to use config for.                                  │
│ --trusted-root      -t  TEXT  A metadata URL to the initial trusted root.                      │
│ --metadata-url      -   TEXT  A URL to download TUF metadata from.                             │
│ --artifacts-url     -a  TEXT  An artifacts base URL to fetch from.                             │
│ --hash-prefix       -h        A hash prefix.                                                   │
│ --directory-prefix  -P  TEXT  A prefix for the download dir.                                   │
│ --help                        Show this message and exit.                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────╯
```

To implement it we add the decorator 
```python
@click.argument('artifact')
@click.pass_context
def download(
    context: Context,
    repository: str,
    trusted_root: str,
    metadata_url: str,
    artifacts_url: str,
    hash_prefix: bool,
    directory_prefix: str,
    artifact,
) -> None:
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+                "artifacts_url"
+            ]
+        if not metadata_url:
+            metadata_url = rstuf_config["repositories"][repository][
+                "metadata_url"
+            ]
+        if not hash_prefix:
+            hash_prefix = rstuf_config["repositories"][repository][
+                "hash_prefix"
+            ]
```
kairoaraujo, line 136: It should be complete optional
```suggestion
        if not hash_prefix:
            hash_prefix = rstuf_config["repositories"][repository].get(
                "hash_prefix", False
            )
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+    Downloads artifacts to the TUF metadata using python-tuf ``ngclient``.
+    """
+
+    rstuf_config: Dict[str, Any]
+    if config:
+        with open(config, "r") as stream:
+            try:
+                rstuf_config = yaml.safe_load(stream)
+            except yaml.YAMLError as e:
+                console.print(f"Failed to parse config file: {e}")
```
kairoaraujo, line 118: We don't need to handle it. 
We use [Dynaconf](https://www.dynaconf.com) and we manage it within config in the `rstuf` command level as I mentioned above.
```suggestion
        rstuf_config = context.obj["settings"]
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+                "metadata_url"
+            ]
+        if not hash_prefix:
+            hash_prefix = rstuf_config["repositories"][repository][
+                "hash_prefix"
+            ]
+        if not trusted_root:
+            trusted_root = rstuf_config["repositories"][repository][
+                "trusted_root"
+            ]
```
kairoaraujo, line 140:
```suggestion
        trusted_root = rstuf_config["repositories"][repository].get("trusted_root")
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+        directory_prefix,
+    )
+
+    if ok:
+        console.print(
+            f"Successfully completed artifact download: {artifact_name}"
+        )
+
+
+def update_artifact_url_with_hash_prefix(artifact_url: str) -> str:
```
kairoaraujo, line 166: nit: Move it to the top

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+    return f"{Path.home()}/.local/share/rstuf/{name}"
+
+
+def init_tofu(metadata_url: str, root_url: Optional[str]) -> bool:
+    """Initialize local trusted metadata (Trust-On-First-Use) and create a
+    directory for downloads"""
+    metadata_dir = build_metadata_dir(metadata_url)
+
+    if not os.path.isdir(metadata_dir):
+        os.makedirs(metadata_dir)
```
kairoaraujo, line 190:
```suggestion
     os.makedirs(metadata_dir, exist_ok=True)
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+import yaml
+from click import Context
+from tuf.api.exceptions import DownloadError, RepositoryError
+from tuf.ngclient import Updater
+
+from repository_service_tuf.cli import click, console
+from repository_service_tuf.cli.artifact import artifact
+from repository_service_tuf.cli.artifact.repository import find_hash_bin
+
+DOWNLOAD_DIR = "./downloads"
```
kairoaraujo, line 22: IMO it shouldn't be nested, it should be the current dir `os. getcwd()` (multiplatform)

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+
+    print(f"Using trusted root in {metadata_dir}")
+
+    if directory_prefix:
+        download_dir = f"{directory_prefix}/downloads"
+    else:
+        download_dir = DOWNLOAD_DIR
+
+    if not os.path.isdir(download_dir):
+        os.mkdir(download_dir)
```
kairoaraujo, line 237: Here is my suggestion:
- The user could have a `download_dir` in the `yml` as optional
- The user can always use the `--download-prefix` (and it overwrites the configuration)
- If none `download_dir` in yaml or none `--download-prefix` we download the local dir (`os. getcwd()`).

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+            console.print(
+                f"Trusted local root not found in {metadata_url}"
+                "`tofu` was not successful"
+            )
+            return False
+
+    print(f"Using trusted root in {metadata_dir}")
+
+    if directory_prefix:
+        download_dir = f"{directory_prefix}/downloads"
```
kairoaraujo, line 232: We shouldn't create a dir, the user decides the prefix
```suggestion
        download_dir = f"{directory_prefix}/"
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+    else:
+        download_dir = DOWNLOAD_DIR
+
+    if not os.path.isdir(download_dir):
+        os.mkdir(download_dir)
+
+    try:
+        updater = Updater(
+            metadata_dir=metadata_dir,
+            metadata_base_url=f"{metadata_url}/metadata/",
```
kairoaraujo, line 242: We shouldn't append `metadata`, for example, https://metadata.dev.rstuf.org/ doesn't use the subfolder.

repository_service_tuf/cli/artifact/repository.py
kairoaraujo: I really don't understand this repository here :)

kairoaraujo (commented):
Hi @ivanayov, thank you for this great work :tada:

I will add some design comments 😄

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,266 @@
...
+    "--config",
+    help="A config file to use.",
+    type=str,
+    required=None,
+    default=None,
+)
+@click.option(
+    "-r",
+    "--repository",
+    help="A repository to use config for.",
```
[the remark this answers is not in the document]
kairoaraujo: IMO, We could remove the `--repository/-r`.
The user should use later the `rstuf artifactory repository use REPOSITORY_NAME`
Optionally, they can use all parameters. 

```shell
│ --trusted-root      -t  TEXT  A metadata URL to the initial trusted root.                      │
│ --metadata-url      -m   TEXT  A URL to download TUF metadata from.                             │
│ --artifacts-url     -a  TEXT  An artifacts base URL to fetch from.                             │
```

ivanayov:
@kairoaraujo @MVrachev Could you please review the updated tests?

docs/source/guide/index.rst
```diff
@@ -913,6 +913,28 @@ This command adds the provided artifact to the TUF Metadata using the RSTUF REST
...
+
+This command allows downloading an artifact from a provided repository using the RSTUF REST API.
+
+.. code::
+
+    > rstuf artifact download --help
+
+    Usage: rstuf artifact download [OPTIONS] ARTIFACT_NAME
+
+    Downloads artifacts to the TUF metadata.
```
MVrachev, line 928: Please add a line here that this artifact will be downloaded from a setting named `current_repository`
and with one sentence mention how you can change that repository.
I just want to make sure that a user of `rstuf artifact download` understands both where he is downloading from and how to change that.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,254 @@
...
+            sys.exit()
+        if not rstuf_config["REPOSITORIES"].get(repository):
+            console.print(
+                f"Repository {repository} is missing in the "
+                "configuration file"
+            )
+            sys.exit()
+        if not artifacts_url:
+            artifacts_url = rstuf_config["REPOSITORIES"][repository].get(
+                "artifact_base_url"
```
MVrachev, line 215: This attribute you want to get is not the same as the one you define in #459.
In #459 the name is `artifact_url`.
I think it's best if we name this `artifacts_base_url` meaning this is not a base url for 1 artifact but many.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,254 @@
...
+    directory_prefix: str,
+    artifact_name,
+) -> None:
+    """
+    Downloads artifacts to the TUF metadata.
+    """
+
+    rstuf_config = context.obj.get("settings")
+
+    if context.obj.get("config") and os.path.isfile(context.obj.get("config")):
```
MVrachev, line 197: What happens if this initla `if` is not true?
I don't see any warnings about this and it will result with `AttributeError` on line 243.

In your place, I would rather do:
```suggestion
    if not context.obj.get("config") or not os.path.isfile(context.obj.get("config")):
        raise click.ClickException("Please specify current repository")
```
and I won't think about this case anymore and all `if` statements will shift to the left.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,254 @@
...
+        repository = rstuf_config.get("CURRENT_REPOSITORY")
+        # Overwriting the config if a flag is passed
+        # or using it as default instead
+        if not repository:
+            console.print("Please specify current repository")
+            sys.exit()
```
MVrachev, line 203: Currently, when we have any error that causes a problem that we cannot resolve we throw an exception.
You can have a look at: https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/main/repository_service_tuf/cli/admin/ceremony.py, https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/main/repository_service_tuf/cli/admin/metadata.py, etc.

I think that will make for shorter code (as you don't need a `sys.exit()` after that.
Also, with this method you can easily forgot the `sys.exit()` call.

Finally, `sys.exit()` exits with a `0` code which is translated to a success.
That's not good especially if it's put in an automation somewhere.
I think you should change the uses of `sys.exit()` here in this code to exceptions to better show there is an actual problem that stops the execution.
Any thoughts @kairoaraujo?

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,254 @@
...
+    hash_prefix = find_hash_bin(artifact_file_name)
+    new_artifact_file_name = hash_prefix + "-" + artifact_file_name
+    splitted = artifact_name.rsplit("/", 1)
+    if len(splitted) > 1:
+        return artifact_name.rsplit("/", 1)[0] + "/" + new_artifact_file_name
+    else:
+        return new_artifact_file_name
+
+
+def decode_trusted_root(root) -> str:
```
MVrachev, line 35: What is the type of this `root`?

MVrachev (changes requested):
Have you tested both `rstuf artifact repository` and `rstuf artifact download` together?
Have you tested with a config file with two repositories and changed which repository you use for downloading and then downloading? 

Reviewed only the main code and not tests.

docs/source/guide/index.rst
MVrachev:
```suggestion
    Downloads artifact(s) to the TUF metadata.
```

repository_service_tuf/cli/artifact/download.py
ivanayov: The initial idea of this was that as the CLI is not considered for automation purposes, but rather used interactively as an UX and when a user is missing a required parameter there is no actual error that happened during execution and a message without a failure is enough. I changed my mind though, as non-zero exit is the standard in such cases - coming from the idea that CLI tools exist for automation by default and, even though `rstuf` differs in that aspect, it's better to stick to the standard.

repository_service_tuf/cli/artifact/download.py
ivanayov: `base64` is frequently causing issues with tox typing and this might cause unexpected CI failures in future and so be a blocker for future PRs. This is why I consider it safer to omit - it will anyway fail with a clear message if the type is improper. We talked about putting `any` instead, to just stick to our convention, but reconsidering, I don't think this is a good idea, as it reads as "root can be any", which is wrong. 
Given that the type restriction anyway falls down to the `base64` library without any additional logic from our side, the optimal solution is to just omit.

docs/source/guide/index.rst
```diff
@@ -913,6 +913,28 @@ This command adds the provided artifact to the TUF Metadata using the RSTUF REST
...
+
+    ╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
+    │ --root              -r  TEXT  A metadata URL to the initial trusted root or a local file.        │
+    │ --metadata-url      -m  TEXT  TUF Metadata repository URL.                                       │
+    │ --artifacts-url     -a  TEXT  An artifacts base URL to fetch from.                               │
+    │ --hash-prefix       -p        A hash prefix.                                                     │
+    │ --directory-prefix  -P  TEXT  A prefix for the download dir.                                     │
+    │ --help              -h        Show this message and exit.                                        │
+    ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```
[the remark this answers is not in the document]
ivanayov: Thanks for noticing! The PR description is outdated, while this addresses a following discussion about the design.

repository_service_tuf/cli/artifact/download.py
ivanayov: It's not considered to fail or raise exceptions. What we do in this `if` (clarified in a comment as well) is to only overwrite the existing config with the corresponding flags when they are passed. What happens if it's False - the user is just not using a config file, so nothing to overwrite.

The reason there's no `ClickException` is that `CURRENT_REPOSITORY` is only a configuration requirement and designed for such use-cases. It's not supported by the prompt, as if you don't use a config file, you specify only a single repository details to refer to. With a config you can list as many repositories, as you wish, but you must point to the one you're using.

repository_service_tuf/cli/artifact/download.py
ivanayov: I don't mind unifying, even though `artifact_url` is for flag simplicity, while `artifact_base_url` is for config clarity. We support both use-cases, but they differ. I agree that `artifacts_base_url` is good for the config file, but it's improper for a flag, as we don't support downloading multiple artifacts at once.

This is why IMO it's best to have both `artifact_url` (flags only) and `artifacts_base_url` (config file). If we choose one, `artifact_url` is best for the command line, even though it's not clear enough for the config file.

docs/source/guide/index.rst
ivanayov: `current_repostitory` is only referred to when an rstuf config is used. May be you mean adding few lines about how to add `artifacts` configuration.

docs/source/guide/index.rst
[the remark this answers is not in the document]
MVrachev: It's hard to remember to update pr description, that's why I mention not going into such details there. :))

repository_service_tuf/cli/artifact/download.py
MVrachev: Okay, I understand your idea now.
Let's leave it like that.

repository_service_tuf/cli/artifact/download.py
MVrachev: That's something worth discussing with @kairoaraujo and @KAUTH as well.
How should we handle a situation where we don't have an argument we need and we cannot continue?
Do we raise an exception and return non-zero code or leave a warning message and exit?

repository_service_tuf/cli/artifact/download.py
MVrachev: I missed that you handle the required attributes in `download_artifact` and that's why I thought you should handle the case the config is missing as well.
I left some comments there.

MVrachev (changes requested):
I left a few more comments.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,263 @@
...
+    if not os.path.isdir(download_dir):
+        os.mkdir(download_dir)
+
+    try:
+        updater = Updater(
+            metadata_dir=metadata_dir,
+            metadata_base_url=metadata_url,
+            target_base_url=artifacts_url,
+            target_dir=download_dir,
+        )
```
MVrachev, line 124: Currently, python-tuf adds sha256 string as a prefix to the artifact name.
There is a configuration object for that based on `UpdaterConfig` object from python-tuf 
that we can use `prefix_targets_with_hash`: https://github.com/theupdateframework/python-tuf/blob/bbe2ca84a9ccd1397ab3229ada2ac2154d60ddad/tuf/ngclient/config.py#L35.
```suggestion
        updater = Updater(
            metadata_dir=metadata_dir,
            metadata_base_url=metadata_url,
            target_base_url=artifacts_url,
            target_dir=download_dir,
            config=UpdaterConfig(prefix_targets_with_hash=False)
        )
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,263 @@
...
+    splitted = artifact_name.rsplit("/", 1)
+    if len(splitted) > 1:
+        return artifact_name.rsplit("/", 1)[0] + "/" + new_artifact_file_name
+    else:
+        # Annotating as it's covered, but codecov doesn't
+        # recognize that properly
+        return new_artifact_file_name  # pragma: no cover
+
+
+def decode_trusted_root(root) -> str:
```
MVrachev, line 37: Do we finally make it `Any`?

tests/files/artifact_download/root.json
MVrachev, line 1: Why do we need this file?

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,263 @@
...
+            "Trust-On-First-Use or copy trusted root metadata to "
+            f"{metadata_dir}/root.json"
+        )
+        ok = init_tofu(metadata_url, root)
+        if not ok:
+            console.print(
+                f"Trusted local root not found in {metadata_url} - "
+                "`tofu` was not successful"
+            )
+            return False
```
MVrachev, line 104: We discussed that instead of `console.print` + `sys.ext()` we would use the traditional way to fail with an exception https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/cda764fec34c2f3371f43e72c7509d941b350a49/repository_service_tuf/cli/admin/import_artifacts.py#L174
Please change all instances where you stopped execution using `console.print()` + `sys.exit()` with an exception.

MVrachev (changes requested):
A few more comments.
I hope soon we will be ready.
I tested the functionality locally.
I have NOT reviewed tests yet.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,263 @@
...
+    required=None,
+    default=None,
+)
+@click.option(
+    "-p",
+    "--hash-prefix",
+    help="A hash prefix.",
+    is_flag=True,
+    required=None,
+)
```
MVrachev, line 180: Most likely there is a mistake with this flag as python-tuf already finds the precise bin for an artifact without the need of us to calculate its bin.

repository_service_tuf/cli/artifact/download.py
ivanayov: base64 library cannot use a parameter, encoded as `any`

tests/files/artifact_download/root.json
ivanayov: We use its contents for some of the test scenarios.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        ok = init_tofu(metadata_url, root)
+        if not ok:
+            console.print(
+                f"Trusted local root not found in {metadata_url} - "
+                "`tofu` was not successful"
+            )
+            return False
+
+    print(f"Using trusted root in {metadata_dir}")
```
kairoaraujo, line 134:
```suggestion
    console.print(f"Using trusted root in {metadata_dir}")
```

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+
+    if directory_prefix:
+        # Annotating as it's covered, but codecov doesn't
+        # recognize that properly
+        download_dir = directory_prefix  # pragma: no cover
+    else:
+        download_dir = os.getcwd() + "/downloads"
+
+    if not os.path.isdir(download_dir):
+        os.mkdir(download_dir)
```
kairoaraujo, line 144:
```suggestion
        os.mkdir(download_dir, exist_ok=True)
```

kairoaraujo (changes requested):

repository_service_tuf/cli/artifact/download.py
ivanayov: Ah, this parameter is still not fixed and causing issues ever since some of the 3.4 versions till now. Now I remember why I removed it.

Breaking tax locally and in the CI. I've put an ignore annotation. If you think this hides risk of unexpected issues at some point, I'd just remove it.

ivanayov: I removed the change, it's still breaking the CI.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+            )
+        if not root:
+            root = rstuf_config["REPOSITORIES"][repository].get("trusted_root")
+            if not root:
+                console.print(
+                    "Trusted root is not cofigured. "
+                    "You should either add it to your config file, "
+                    "or use the download commang without a config file"
+                )
+                return
```
MVrachev, line 258: This should be an exception.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,263 @@
...
+        f"Trust-on-First-Use: Initialized new root in {metadata_dir}"
+    )
+    return True
+
+
+def download_artifact(
+    metadata_url: Optional[str],
+    artifacts_url: Optional[str],
+    artifact_name: str,
+    root: Optional[str],
```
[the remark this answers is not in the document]
MVrachev: `root` is still `Optional` when it shouldn't be.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+        console.print(f"Failed to download initial root from {root_url}")
+        return False  # pragma: no cover
+
+    console.print(
+        f"Trust-on-First-Use: Initialized new root in {metadata_dir}"
+    )
+    return True
+
+
+def perform_tuf_ngclient_download_artifact(
```
MVrachev, line 61: This is a helper function.
The name should start with `_`.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+
+from click import Context
+from tuf.api.exceptions import DownloadError, RepositoryError
+from tuf.ngclient import Updater, UpdaterConfig
+
+from repository_service_tuf.cli import click, console
+from repository_service_tuf.cli.artifact import artifact
+
+
+def decode_trusted_root(root) -> str:
```
MVrachev, line 22: As this is a helper function please add a `_` prefix.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+    root_encoded = bytes(root, "utf-8")
+    return base64.b64decode(root_encoded).decode("utf-8")
+
+
+def build_metadata_dir(metadata_url: str) -> str:
```
MVrachev, line 27: As this is a helper function please add a `_` prefix.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+    """build a unique and reproducible directory name for the repository url"""
+    name = sha256(metadata_url.encode()).hexdigest()[:8]
+    return f"{Path.home()}/.local/share/rstuf/{name}"
+
+
+def init_tofu(metadata_url: str, root_url: Optional[str]) -> bool:
```
MVrachev, line 33: As this is a helper function please add a `_` prefix.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+            return True
+
+        path = updater.download_target(info)  # pragma: no cover
+        console.print(  # pragma: no cover
+            f"Artifact downloaded and available in {path}"
+        )
+
+    except (OSError, RepositoryError, DownloadError) as e:
+        console.print(f"Failed to download artifact {artifact_name}: {e}")
+        return False
```
MVrachev, line 98: This should be an exception instead.
We don't need to return anything from this function. It doesn't make any boolean check and if it didin't raise anything it succeded.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+            "Trust-On-First-Use or copy trusted root metadata to "
+            f"{metadata_dir}/root.json"
+        )
+        ok = init_tofu(metadata_url, root)
+        if not ok:
+            console.print(
+                f"Trusted local root not found in {metadata_url} - "
+                "`tofu` was not successful"
+            )
+            return False
```
MVrachev, line 132: It's okay that you logged inside `in_tofu` the error itself and returned a boolean, but here I think you should raise an exception.
Logging from a function and postponing raising an exception is useful to give more context to the user, but when there is no context left to be added just raise the exception with the message from the `console.print`.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+    res = perform_tuf_ngclient_download_artifact(
+        metadata_url,
+        metadata_dir,
+        artifacts_url,
+        artifact_name,
+        download_dir,
+        config,
+    )
+
+    return res
```
MVrachev, line 159: No need of parsing any boolean result as I mentioned above.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -0,0 +1,274 @@
...
+        hash_prefix,
+        directory_prefix,
+        artifact_name,
+        root,
+    )
+
+    if ok:
+        console.print(
+            f"Successfully completed artifact download: {artifact_name}"
+        )
```
MVrachev, line 274: No need of parsing any boolean result as I mentioned above.
If it reached the next line after `_download_artifact` its clear it didn't raise.

MVrachev (changes requested):

repository_service_tuf/cli/artifact/download.py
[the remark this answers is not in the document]
ivanayov: We can have an already existing `root.json` key, in which case we don't use the parameter, or we only use it for `init_tofu()` otherwise, where we check the parameter and retrieve a key if it's missing.

repository_service_tuf/cli/artifact/download.py
```diff
@@ -120,7 +120,7 @@ def _download_artifact(
             "Trust-On-First-Use or copy trusted root metadata to "
             f"{metadata_dir}/root.json"
         )
-        _init_tofu(metadata_url, root)
+        _init_tofu(metadata_url, root_url)
```
ivanayov, line 123: Same here

tests/conftest.py
```diff
@@ -183,3 +184,12 @@ def metadata_sign_input() -> List[str]:
     ]
 
     return input
+
+
+@pytest.fixture()
+def mocked_os_makedirs(monkeypatch):
```
ivanayov, line 190: Given it's used only for the `repository_service_tuf.cli.artifact.download.os.makedirs`, isn't it better to be in the corresponding `test_download.py`?

MVrachev (changes requested):
I have three final changes that I see that need to be made.
Otherwise, great work!

MVrachev:
I allowed myself to remove this new root file you were adding and instead the one we already have in: https://github.com/repository-service-tuf/repository-service-tuf-cli/tree/main/tests/files.
I also fixed on error message.

What is left is to only separate tests with config and without and for you to review my changes.

ivanayov:
@MVrachev I reviewed your latest updates - looks ok. Pushed the latest tests change with separating classes for better readability, so would be happy if you can review :) Thanks!

MVrachev (approved):

kairoaraujo (approved):

PR #447, opened 2023-11-10
Fix expected message when calling sign delete

MVrachev:
The expected message was different than the one send from the API.

Also, make sure we are testing how we call "send_payload" and "task_status".

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>

# Code of Conduct


MVrachev:
Here is exactly what happens when you call `rstuf admin metadata sign --delete` after DAS bootstrap is completed:
image

We need to be careful and make sure we always manually test the features besides adding unit tests.

tests/unit/cli/admin/test_metadata.py
```diff
@@ -1194,3 +1194,19 @@ def test_metadata_sign_delete(self, client, test_context):
...
+                "Metadata delete sign",
+            )
+        ]
+        assert metadata.task_status.calls == [
+            pretend.call(
+                "task_id",
+                test_context["settings"],
+                "Metadata delete sign",
+            )
+        ]
```
MVrachev, line 1212: @ivanayov please make sure that on each function you use `pretend` you also check its calls as I did here.

repository_service_tuf/cli/admin/metadata.py
```diff
@@ -729,12 +729,12 @@ def sign(context, api_server: Optional[str], delete: Optional[bool]) -> None:
             settings,
             URL.METADATA_SIGN_DELETE.value,
             payload,
-            "Metadata delete sign accepted.",
+            "Metadata sign delete accepted.",
             "Metadata delete sign",
         )
         task_status(task_id, settings, "Signing process status: ")
         console.print("\nSigning process deleted!\n")
-        sys.exit()
```
MVrachev, line 737: Not sure what is the benefit of using `sys.exit()` as compared to return...

kairoaraujo (approved):

PR #470, opened 2023-12-14
Update securesystemslib and make the necessary changes

MVrachev:
Update securesystemslib + other dependencies and make the necessary changes.

We want to see why are the functional tests failing at issue #372 as the logs are lost.

Fixes #372

PR #477, opened 2023-12-21
Add alternative sign cli

lukpueh:
Invoke as `rstuf admin2 sign`.

**Advantages compared to existing cli**

- asks user to choose signing key from relevant keys only
- gives comprehensive user feedback about signature status
- only uses state-of-the-art securesystemslib interfaces (no legacy)
- simpler implementation design (no redundant Metadata abstraction)
- 'load and sign with relevant keys and give feedback' subroutine is designed for re-use in ceremony and metadata update

**TODO**
- API integration
  - pull metadata
  - push metadata
  - stop signing event
- Improve wording and visual (for review before sign)


lukpueh:
FYI, if you want to try this out: The docstrings in the added test file show the output of complete runs for signing the added test root metadata v1 and v2 respectively, using the added test keys. 

(make sure to have the latest securesystemslib installed in your environment)

MVrachev:
Before considering this pr I think we need https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/470

MVrachev:
Please, rebase on top of #470.

lukpueh:
I rebased and pushed a small docstring fix. Lint job fails because of wip test stubs, please ignore, they need to be changed anyway.

I think this is ready for a high-level review. E.g. try out the dialog as suggested in https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/477#issuecomment-1866313745 and skim the code if it looks sane enough to continue working on it.

In the meanwhile, I'll add ceremony and update to admin2.

MVrachev:
I will try and do that review on phases.

First, phase for me is `Structure`, `Functionality` and `Flow`.
Then, I will focus on UI/UX and finally on the code.

## Structure
1. I prefer if we follow a similar paths as before meaning we can separate some of the commands in a separate module.
In this case we can have `repository_service_tuf/cli/admin/metadata/sign2.py` module for all sign commands and `repository_service_tuf/cli/admin/metadata/metadata2.py` for other metadata commands
Alternatively we can combine metadata and sign commands into one: ``repository_service_tuf/cli/admin/metadata2.py` file as it's now.
2. I preffer if we use full names for our keys based on their type meaning:
- `ec` -> `ecdsa`
- `ed` -> `ed25519`
etc.
3. Rename `tests/files/root/v1.json` and `tests/files/root/v2json` parent folders to be clear from the naming of those files that they need signatures. Maybe add something like `unsigned` or` untrusted` to the naming scheme. 

## Flow
I will do a comparison betwen the flow in my (from this branch: https://github.com/MVrachev/repository-service-tuf-cli/tree/prepare-md-update) and yours:

`My pr flow`:

1. Choose metadata to sign
2. Always print metadata (given that it’s a table and its small we can do that as compared to JSON preview)
3. Do you still want to sign <METADATA_ROLE>?
4. Show pending keys
    1. if metadata_to_sign.version < 2 ⇒ print pending keys only for metadata_to_sign
    2. else ⇒ print pending keys for metadata_to_sign AND trusted_root pending keys
5. choose a key to sign from all possible pending keys
6. sign
7. done

`Yours`:

1. Enter path to sign
2. If root_loaded.version > 2: enter previous root path
3. needed X signatures from root to sign...
    1. if previous root loaded ⇒ needed Y signatures from previous root
4. Review?
5. Sign?
6. Choose key
7. Enter path to private key file
8. Enter password
9. Verification check:
    1. If metadata not verified go to 3.
    2. else done.

**A few notes:**
1. You don't allow partial signing - meaning signing with only 1 key and leaving the metadata file in DAS state.
2. I think its better if you first ask the user if he wants to review the metadata file and THEN show the required signatures.
3. It's good you support signing with multiple keys

## Features

For now I will just write down summarized all the features I noticed that are missing including what you mentioned above (this could be updated):
1. pull metadata
2. push metadata
3. Make the questions independent from the role. Don’t mention `root`, but after getting the pending roles from the repository you can ask which metadata should be sign. This will be useful when we start supporting offline targets as we plan.
4. `rstuf admin metadata sign --delete` used to delete/stop a DAS process. Should be replaced with something else.

lukpueh:
Thanks for the detailed review, @MVrachev!

> ## Structure

Your suggestions are reasonable. Is it okay, to make those changes in the very end? Right now we haven't even decided, if we want to keep this CLI, or just use it as inspiration for the existing CLI. Shuffling code around seems premature.

> ## Flow
> 1. You don't allow partial signing - meaning signing with only 1 key and leaving the metadata file in DAS state.

Actually, I do. When the user is asked to "Sign? [y/n]", they can exit the loop even if the threshold is not reached yet.

> 2. I think it's better if you first ask the user if he wants to review the metadata file and THEN show the required signatures.

May I ask, why you think it's better? There are two reasons, why I chose this order:

1. The signature check should definitely happen first, because there is no reason to ask the user if they want to review when the metadata is already fully signed. Of course, I could only check and still ask about review first.
2. However, I think it makes sense to show the user the missing signatures first regardless. Because it could show them that their signature is not missing. Then it also doesn't make sense for them to review.

> ## Features 
> 1. pull metadata
> 2. push metadata

👍

> 3. Make the questions independent from the role. Don’t mention `root`, but after getting the pending roles from the repository you can ask which metadata should be sign. This will be useful when we start supporting offline targets as we plan.

Right now this is a tool to sign root metadata. Doesn't it make sense if the UI reflects that?

> 4. `rstuf admin metadata sign --delete` used to delete/stop a DAS process. Should be replaced with something else.

👍 Yeah, that's part of the missing API integration. Maybe `delete` could be an independent command.

MVrachev:
> > ## Structure
> 
> Your suggestions are reasonable. Is it okay, to make those changes in the very end? Right now we haven't even decided, if we want to keep this CLI, or just use it as inspiration for the existing CLI. Shuffling code around seems premature.

Yes, that's fine for me.

> 
> > ## Flow
> > 
> > 1. You don't allow partial signing - meaning signing with only 1 key and leaving the metadata file in DAS state.
> 
> Actually, I do. When the user is asked to "Sign? [y/n]", they can exit the loop even if the threshold is not reached yet.

Yes, I missed that.
 
> > 2. I think it's better if you first ask the user if he wants to review the metadata file and THEN show the required signatures.
> 
> May I ask, why you think it's better? There are two reasons, why I chose this order:
> 
> 1. The signature check should definitely happen first, because there is no reason to ask the user if they want to review when the metadata is already fully signed. Of course, I could only check and still ask about review first.

As you said, nothing stops you from checking the signatures before the metadata preview.

> 2. However, I think it makes sense to show the user the missing signatures first regardless. Because it could show them that their signature is not missing. Then it also doesn't make sense for them to review.

Overall, I think I will do this part differently as I did in my flow.
Here is how:
1. I think we should directly show the metadata file after giving the path.
It's just logical to me to do that directly after you have given the path with a sentence before that `Here is the content of that file:`. 
2. I wouldn't ask the user if he wants to review the file. This operation is important to not mess up and given that it will be done rarely I don't see why we would ask the user.
Also, I believe that we would (should) beautify the way we show our metadata file and it will be more succinct.
3. After the preview I would add a question:
`Do you want to sign <METADATA_ROLE_NAME_HERE>`.
This would be useful to remind the maintainer he needs to make sure he agrees with the metadata content.
5. After the user agrees I would show him a table with pending keys with a sentence:
`<METADATA_ROLE_NAME_HERE> role still needs X numbers of keys from <KEY_NAMES>`
6. If there a previous trusted root, I will print the same table and sentence as on step 5, but for trusted root pending keys.

To summarize, what I would do as an end-flow is the folling:
1. Enter path to sign
2. If `das_root_version > 2` =>  enter previous root path
3. `Here the content of the file to sign` + Show the content of that file
4. Ask `Do you want to sign <METADATA_ROLE_NAME_HERE>?`
5. If `das_root.version > 2` => show key table with `trusted_root pending keys` + sentence:
`<METADATA_ROLE_NAME_HERE> role still needs X numbers of keys from <KEY_NAMES>`
6. Show key table with `das_root` pending keys` + sentence:
`<METADATA_ROLE_NAME_HERE> role still needs X numbers of keys from <KEY_NAMES>`
7. Choose key (from all pending keys comming from das and trusted roots)
8. Enter path to private key file
9. Enter password
10. Verification check:
11. If metadata not verified go to 2.
12. else done.

> > ## Features
> > 
> > 1. pull metadata
> > 2. push metadata
> 
> 👍

I just want to mention here that when we pull metadata we would also pull the latest trusted root version, not just the DAS root.
I implemented that as a preparation for DAS signing:
https://github.com/repository-service-tuf/repository-service-tuf-api/pull/499

> 
> > 3. Make the questions independent from the role. Don’t mention `root`, but after getting the pending roles from the repository you can ask which metadata should be sign. This will be useful when we start supporting offline targets as we plan.
> 
> Right now this is a tool to sign root metadata. Doesn't it make sense if the UI reflects that?

Now, that's correct.
We plan that this won't by a tool only for root, but this is a low priroty note.
 
> > 4. `rstuf admin metadata sign --delete` used to delete/stop a DAS process. Should be replaced with something else.
> 
> 👍 Yeah, that's part of the missing API integration. Maybe `delete` could be an independent command.

I would add one more: when asking the user which key he wants to use don't show keyids, but instead show key names.

lukpueh:
> 1. I think we should directly show the metadata file after giving the path.
>    It's just logical to me to do that directly after you have given the path with a sentence before that `Here is the content of that file:`.

Okay, but what about my argument, that the user might not even have the right key?  Wouldn't it be mean to let the user review the thing and afterwards tell them that they can't sign it anyway?


> 2. I wouldn't ask the user if he wants to review the file. This operation is important to not mess up and given that it will be done rarely I don't see why we would ask the user.

My idea was to allow a flow with minimal output, which focuses on the signing and key handling business. Because for review, the user might prefer a different tool (e.g. `diff`). But I'm fine either way.

>    Also, I believe that we would (should) beautify the way we show our metadata file and it will be more succinct.

Agreed! We shouldn't make anyone read json.

> 3. After the preview I would add a question:
>    `Do you want to sign <METADATA_ROLE_NAME_HERE>`.
>    This would be useful to remind the maintainer he needs to make sure he agrees with the metadata content.

Yeah, that's basically my "Sign?" prompt. We can definitely improve the wording.

> 4. After the user agrees I would show him a table with pending keys
>    Here is an example key table that I made:
> 
> image

I do think the table is pretty, but I would also say that it has a lot of information that the user doesn't really care for. I would establish a single identifier, by which the user knows what key to use.

MVrachev:
> > 1. I think we should directly show the metadata file after giving the path.
> >    It's just logical to me to do that directly after you have given the path with a sentence before that `Here is the content of that file:`.
> 
> Okay, but what about my argument, that the user might not even have the right key? Wouldn't it be mean to let the user review the thing and afterwards tell them that they can't sign it anyway?

What will be the use case when a user wouldn't have the key?
If a root DAS process underway then it will make that only root maintainers will care about the `/api/v1/metadata/sign` API calls. Who will use this API call and not have a key used in root?
The only real scenario I imagine if you already signed it and forgot, but I see this as a corner case.

My strong feeling is that we should always show the users the root metadata they sign to make sure they are sure what they are doing.
The debate I think is when to show it.

> > 2. I wouldn't ask the user if he wants to review the file. This operation is important to not mess up and given that it will be done rarely I don't see why we would ask the user.
> 
> My idea was to allow a flow with minimal output, which focuses on the signing and key handling business. Because for review, the user might prefer a different tool (e.g. `diff`). But I'm fine either way.

Yea, It's a good argument, but considering this will be done rarely I don't see it as big problem.
Maybe we need to think wouldn't be better if we could show a diff somehow in a nice way?
A diff in a table? :D 

> > Also, I believe that we would (should) beautify the way we show our metadata file and it will be more succinct.
> 
> Agreed! We shouldn't make anyone read json.
> 
> > 3. After the preview I would add a question:
> >    `Do you want to sign <METADATA_ROLE_NAME_HERE>`.
> >    This would be useful to remind the maintainer he needs to make sure he agrees with the metadata content.
> 
> Yeah, that's basically my "Sign?" prompt. We can definitely improve the wording.

Oh, yes, correfct.

> > 4. After the user agrees I would show him a table with pending keys
> >    Here is an example key table that I made:
> > 
> > image
> 
> I do think the table is pretty, but I would also say that it has a lot of information that the user doesn't really care for. I would establish a single identifier, by which the user knows what key to use.

Well, my way of thinking was that I want to give the user enough context that he understands which of his (probably more than 1) offline keys is.
That's why I also added key type and key value.
What we can improve is make it such that either key name or key id is shown as if the key name is given there is no point of key id, I think? 
KeyId will be used if the name is missing and the user needs to see more information about a key mentioned in the prompt: 
`Choose key (from all pending keys comming from das and trusted roots)`.

lukpueh:
> > Okay, but what about my argument, that the user might not even have the right key? Wouldn't it be mean to let the user review the thing and afterwards tell them that they can't sign it anyway?
> 
> What will be the use case when a user wouldn't have the key? If a root DAS process underway then it will make that only root maintainers will care about the `/api/v1/metadata/sign` API calls. Who will use this API call and not have a key used in root? The only real scenario I imagine if you already signed it and forgot, but I see this as a corner case.

Sure, but I think the tool is not limited to be used with the API, right? Regardless of the scenario, don't you agree that it's nicer UX to first ask an easy question (*Do you have one of those singing keys?*), before you ask a hard question (*Are you sure that those verification keys should be the root of trust?*), if the answer to the easy question might make the hard question irrelevant? :)


> 
> My strong feeling is that we should always show the users the root metadata they sign to make sure they are sure what they are doing. The debate I think is when to show it.

I think we agree on that. :) 

> > > image
> > 
> > 
> > I do think the table is pretty, but I would also say that it has a lot of information that the user doesn't really care for. I would establish a single identifier, by which the user knows what key to use.
> 
> Well, my way of thinking was that I want to give the user enough context that he understands which of his (probably more than 1) offline keys is. That's why I also added key type and key value. What we can improve is make it such that either key name or key id is shown as if the key name is given there is no point of key id, I think? KeyId will be used if the name is missing and the user needs to see more information about a key mentioned in the prompt: `Choose key (from all pending keys comming from das and trusted roots)`.

Yeah, maybe, but:
- why mention that the key is "offline"?
- is it likely that a user distinguishes their keys by key type? isn't it just as likely that the user only has ed25519 keys?
- how does a public key value in some (to me) unknown encoding help me to pick a private key file on disk (or a yubikey in the future)?

MVrachev:
> > > Okay, but what about my argument, that the user might not even have the right key? Wouldn't it be mean to let the user review the thing and afterwards tell them that they can't sign it anyway?
> > 
> > 
> > What will be the use case when a user wouldn't have the key? If a root DAS process underway then it will make that only root maintainers will care about the `/api/v1/metadata/sign` API calls. Who will use this API call and not have a key used in root? The only real scenario I imagine if you already signed it and forgot, but I see this as a corner case.
> 
> Sure, but I think the tool is not limited to be used with the API, right? Regardless of the scenario, don't you agree that it's nicer UX to first ask an easy question (_Do you have one of those singing keys?_), before you ask a hard question (_Are you sure that those verification keys should be the root of trust?_), if the answer to the easy question might make the hard question irrelevant? :)

Okay, let's add such a question. I just don't want to mention key requirements at this stage.
I prefer if we show them in a table view of some kind as I proposed above.

> > My strong feeling is that we should always show the users the root metadata they sign to make sure they are sure what they are doing. The debate I think is when to show it.
> 
> I think we agree on that. :)
> 
> > > > image
> > > 
> > > 
> > > I do think the table is pretty, but I would also say that it has a lot of information that the user doesn't really care for. I would establish a single identifier, by which the user knows what key to use.
> > 
> > 
> > Well, my way of thinking was that I want to give the user enough context that he understands which of his (probably more than 1) offline keys is. That's why I also added key type and key value. What we can improve is make it such that either key name or key id is shown as if the key name is given there is no point of key id, I think? KeyId will be used if the name is missing and the user needs to see more information about a key mentioned in the prompt: `Choose key (from all pending keys comming from das and trusted roots)`.
> 
> Yeah, maybe, but:
> 
> * why mention that the key is "offline"?

This could be removed, I agree.

> * is it likely that a user distinguishes their keys by key type? isn't it just as likely that the user only has ed25519 keys?

This is a useful field I think. It's possible that not all of your keys are  ed25519.

> * how does a public key value in some (to me) unknown encoding help me to pick a private key file on disk (or a yubikey in the future)?

Isn't the public value encoded in a standard way not specific to securesystemslib? 
If the encoding is not popular, then I agree we can skip it.

I am thinking that having at least `keyid` and key name at the key table becaus:
- `keyid` is useful to make a crossreference with previous root(s) or with the information about the current das root you are about to sign. 
- `key name` is an easy way to reference a key and can be used for custom context provided by the maintainers

Based on our discussion about the flow so far I propose the following flow:
1. Enter path to sign
2. If `das_root_version > 2` => enter previous root path
3. Do you have one of those singing keys? (show names of pending keys)
4. `Here the content of the file to sign` + Show the content of that file
5. `Do you want to sign <METADATA_ROLE_NAME_HERE>?`
6. If `das_root.version > 2` => show key table with `trusted_root pending keys`
7. Show key table with `das_root` pending keys
8. Choose key (from all pending keys comming from das and trusted roots)
9. Enter path to private key file
10. Enter password
11. Verification check:
12. If metadata not verified go to 2.
13. else done.

lukpueh:
> Isn't the public value encoded in a standard way not specific to securesystemslib? If the encoding is not popular, then I agree we can skip it.

I'm saying that the user probably doesn't and definitely shouldn't need to know which encoding it is. Because even if they knew, how are they supposed to identify the corresponding signing key? For instance, how do you pick the right [private key file](https://github.com/repository-service-tuf/repository-service-tuf-cli/tree/2db958ee9334499f474da527e13a0a9f865a1976/tests/files/pem) for a public key value `4f66dabebcf30628963786001984c0b75c175cdcf3bc4855933a2628f0cd0a0f`?

> Based on our discussion about the flow so far I propose the following flow:

> 
> 1. Enter path to sign
> 2. If `das_root_version > 2` => enter previous root path
> 3. Do you have one of those singing keys? (show names of pending keys)
> 4. `Here the content of the file to sign` + Show the content of that file
> 5. `Do you want to sign <METADATA_ROLE_NAME_HERE>?`
> 6. If `das_root.version > 2` => show key table with `trusted_root pending keys`
> 7. Show key table with `das_root` pending keys
> 8. Choose key (from all pending keys comming from das and trusted roots)
> 9. Enter path to private key file
> 10. Enter password
> 11. Verification check:
> 12. If metadata not verified: Do you want to contiue signing?
> 13. If yes go to 3.
> 14. else done.


Yeah, this flow is pretty much what I propose, except that I would show 6/7 in 3 and try to reduce the amount of information needed for a user to identify the signing for a private key. Maybe we can have another brainstorm about it next week.

MVrachev:
Notes from a meeting with @lukpueh and @kairoaraujo.

## Missing features & required changes 
1. Pull metadata from RSTUF API: https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/38b739fba75284a1162e0aa2a9564afca38c2aca/repository_service_tuf/cli/admin/metadata.py#L611
2. Push a signature to `POST /api/v1/metadata/sign`.
- we decided we will allow only 1 signing at a time.

## UI/UX features to add:
1. Add an introduction to the command - I will help
2. Use more colors on prompts to highlight better what is required - I will help
3. A review of the Root metadata should be in a table - I will help
4. New line between root review and keys requirement
5. Show key requirements as a table - I will help.
6. Show choice of a key using key names
7. When we save the file, save it with new spaces - beatify it - Lukas

lukpueh:
> 7. When we save the file, save it with new spaces - beautify it - Lukas

I saw that the current `sign` command does not support files. Should we just remove the option to save to file, once API integration works?

MVrachev:
> > 7. When we save the file, save it with new spaces - beautify it - Lukas
> 
> I saw that the current `sign` command does not support files. Should we just remove the option to save to file, once API integration works?

I am okay with that.

lukpueh:
@MVrachev, I think I addressed most of your comments. The following tasks remain:
- Add an introduction to the command
- Use more colors on prompts to highlight better what is required
- A review of the Root metadata should be in a table

Do you feel like working on those? Also, should we maybe move on to a code review and merge this PR first (it doesn't have to be main, but could be an `admin2` branch or so)? Then you can submit style improvement PRs on top, and I can add tests. cc @kairoaraujo

lukpueh:
Regardless, I'll fix ci in this PR. Not sure why it's failing. Tests do pass locally.

lukpueh:
On a different note, now that the sign cli can't  load local metadata anymore it becomes a bit harder to test this. Here's a quick and dirty way to run the cli, with mocked api responses that return locally loaded metadata. Maybe that's helpful for reviewers.

```python
from unittest.mock import patch
from tuf.api.metadata import Metadata, Root
import repository_service_tuf.cli.admin2 as admin2

REPO_PATH = "."

root1 = Metadata[Root].from_file(f"{REPO_PATH}/tests/files/root/v1.json")
root2 = Metadata[Root].from_file(f"{REPO_PATH}/tests/files/root/v2.json")

with patch.object(
    admin2, "_fetch_metadata", return_value=(root2, root1.signed)
), patch.object(admin2, "_push_signature", return_value=None):

    admin2.sign.main(["--api-server=bla"], standalone_mode=False)
```

lukpueh:
Just force pushed to sync with main. But only the last 4 commits are new. I made some minor improvements and two bugs.

Unfortunately, I couldn't fix the ci failure yet. It is related to the latest `cryptography` update, whose `load_from_private_pem` method doesn't recognise our encrypted test ecdsa and rsa keys.

MVrachev:
I was about to test this, but realized something. 

I just realized that after your changes we can only test the signing if the metadata is retrieved from the API url, but your current example root files `v1.json` and `v2.json` don't have any signatures and thus cannot be used for testing of this command.
Please update them with at least one signature per file.

lukpueh:
> I just realized that after your changes we can only test the signing if the metadata is retrieved from the API url, but your current example root files `v1.json` and `v2.json` don't have any signatures and thus cannot be used for testing of this command.

I don't understand how those things are related. May I ask how you are testing this? The cli currently does not require any signatures on the metadata (I think the original cli doesn't either). So you should be able to test the tool with those two files (see the included smoke test).

IMO the bigger inconvenience since the mentioned change is that now you need to provide the metadata via API. In the test and the [snippet](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/477#issuecomment-1917318704) above I work around this my mocking the methods, which interact with the API.

I'm happy to discuss the test strategy for the cli, but I thought we could do that independently and maybe in a follow-up PR.

lukpueh:
fyi @MVrachev, I just removed some code that we upstreamed to python-tuf, and used the new API instead. For the time being, this means that we have to use a non-released version of python-tuf. But it will be released soon.

repository_service_tuf/cli/admin2.py
```diff
@@ -0,0 +1,277 @@
...
+    # Show only one result, if the same number of signatures from the same set
+    # of keys is missing in both.
+    if (
+        results.second.unsigned != results.first.unsigned
+        or results.second.missing != results.first.missing
+    ):
+        results_to_show.append(results.second)
+
+    for result in results_to_show:
+        title = f"Please add {result.missing} more signature(s) from any of "
```
lukpueh, line 62: Just saw that this can be 
```
Please add 0 more signature(s) from any of 
```

will change to show only unverified results

PR #490, opened 2024-01-19
Add alternative metadata update cli

lukpueh:
*includes commits from https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/477*

Invoke as `rstuf admin2 update`

Provides root metadata update dialog similar to the existing `rstuf admin metadata update`. Users can change expiry date, root keys + threshold, and the online key, and re-sign the updated root metadata.

**Advantages compared to existing tool**
- Better separates public and private key management. This allows metadata update regardless of private key availability (needed for DAS), and sets the stage for other key technologies (e.g. HSM, cloud KMS, etc.)
- Introduces signer URIs, included with the online public key. This makes configuration of arbitrary online signing providers (e.g. cloud KMS) substantially easier (see option 2 in [repository-service-tuf/repository-service-tuf#580](https://github.com/repository-service-tuf/repository-service-tuf/issues/580)).
- Uses only state-of-the-art securesystemslib API
- Re-uses signing dialog from alternative sign cli ([#477](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/477))
- Is designed for re-use in `rstuf admin2 bootstrap`
- Omits redundant data model on top of existing TUF Metadata API model.

**Why not update the existing tool?**
Above all, the strong coupling between public and private key management, tied to the legacy securesystemslib API, would have required rather invasive changes. A rewrite seemed more feasible. That said, I took a lot of inspiration from the existing tool, and highly value all the related efforts.

**Follow-up work**
- API integration
- Polish wording and visual
  - Should be consistent, informative, succinct
  - Most importantly, how much public key detail is needed when?
- Maybe polish flow
- Improve test coverage and tests
- Also see TODO code comments


lukpueh:
Regarding  "Maybe polish flow", here is what I had in mind...

The original cli followed a prompt pattern:
```
Do want to change X? -> [if yes] Change X: -> <change> -> Is X ok? ->  [if no] <repeat>
```
The new cli has streamlined this to:
```
Do you want to change X? -> [if yes] Change X: -> <change and repeat>
```
Maybe this can be further streamlined to:
```
Change X or skip: - [if not skip] <change and repeat>
```

What do others think?

MVrachev:
Comments that were mentioned during a meeting between @lukpueh, @kairoaraujo and me"

1. Ask first to remove or add key and in the end ask about threshold
2. Add more threshold validation:
- Make sure that `threshold < number of keys`
- 
3. We should allow saving the payload, not the file
4. We don’t allow signing a local file in the future 
5. Step `5` from `Do you want to change root keys? [y/n]` to `Do you want to change root keys OR THRESHOLD? [y/n]`
to be clear to the user that he will miss the opportunity to change `threshold` if he answers with `n`.
6. Choose a key to remove: show key names insted of keyids if they exists else default to keyid
11. Make `key name` mandatory when asking the user during ceremony or md update add key
12. Make sure that key name and keyids are unique
13. Step 10 `Please enter a public key path:` we can add loading a private key directly as an option and if so don't ask about private key information for that key during signing.

lukpueh:
> Comments that were mentioned during a meeting between @lukpueh, @kairoaraujo and me"
> 
> 1. Ask first to remove or add key and in the end ask about threshold
> 2. Make sure that `threshold < number of keys`
> 3. We should allow saving the payload, not the file
> 4. We don’t allow signing a local file in the future
> 5. Step `5` from `Do you want to change root keys? [y/n]` to `Do you want to change root keys OR THRESHOLD? [y/n]`
>    to be clear to the user that he will miss the opportunity to change `threshold` if he answers with `n`.
> 6. Choose a key to remove: show key names insted of keyids if they exists else default to keyid
> 7. Make `key name` mandatory when asking the user during ceremony or md update add key
> 8. Make sure that key name and keyids are unique
> 9. Step 10 `Please enter a public key path:` we can add loading a private key directly as an option and if so don't ask about private key information for that key during signing.

lukpueh:
Just pushed plenty of new commits to address most of https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/490#issuecomment-1916750572, 

I will pause this work for now and move on to `rstuf admin2 ceremony`, where I hope to re-use some of the subroutines from here... which I expect to entail some changes here.

Feel free to ignore this PR until I request a review!

lukpueh:
Superseded by #523

PR #503, opened 2024-02-05
Custom Timeout

### LINKED ISSUE #328: Feature: Custom timeout for Ceremony Bootstrap
### What is the task about?

This task is a feature implementation on the CLI to support a custom timeout.
A request to the RSTUF API does the Bootstrap.
The request is done to the `/api/v1/bootstrap` using the method `POST` and it is used by `rstuf admin ceremony --bootstrap`
The endpoint supports a custom timeout (default is 300 seconds).

This issue is about providing the timeout parameter in the CLI when the user uses `-b/—bootstrap`.

### Parent feature

_No response_

### References

API Documentation
- https://repository-service-tuf.github.io/repository-service-tuf-api/#/v1/post_api_v1_bootstrap__post

### Code of Conduct

muhahahmad68:
# Description

Fixes #328 

# Types of changes

# Additional requirements

# Code of Conduct


repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -717,7 +723,7 @@ def ceremony(
     # option upload: it requires -b/--bootstrap
     if upload is True and bootstrap is False:
         raise click.ClickException("Requires '-b/--bootstrap' option.")
-
+
```
MVrachev, line 726: Could you remove the extra space?

tests/unit/cli/admin/test_ceremony.py
```diff
@@ -554,7 +554,7 @@ def test_ceremony_option_save(
         ]
         assert ceremony._run_ceremony_steps.calls == [pretend.call(True)]
         assert ceremony.save_payload.calls == [
-            pretend.call("payload.json", {"k": "v"})
+            pretend.call("payload.json", {"k": "v", "timeout": 300})
         ]
 
     def test_ceremony_option_save_OSError(
```
MVrachev, line 560: Instead of using an old test and adding the timeout in the payload what is better to do is to add a separate test maybe called `test_ceremony_option_timeout` and use that.
As an example, we have a separate test for the `save` option starting on line 518.

muhahahmad68:
If I get you. I'm creating a new function for the timeout, with the same functionality as the function i modified, but with additional argument for timeout?

kairoaraujo (approved):

MVrachev:
> If I get you. I'm creating a new function for the timeout, with the same functionality as the function i modified, but with additional argument for timeout?

You have modified the test function named `test_ceremony_option_save` by adding an additional argument `timeout`.
That works, but it's not following our convention.
What we do is that we have a test function for each separation option.
The test you have modified `test_ceremony_option_save` uses the `save` option.

You need to add a test function named `test_ceremony_option_timeout` and then add only the timeout option.
That makes it clear where do we test each of our options.
 If it's still not clear write me on slack.

MVrachev (approved):

MVrachev (approved):

MVrachev (approved):

MVrachev (approved):

MVrachev (approved):

PR #577, opened 2024-05-10
Bump root md for tests for current admin commands

MVrachev:
Bump root md for tests for current admin commands

This is a temporary fix, so our UT tests will not fail.

Signed-off-by: Martin Vrachev <martin.vrachev@broadcom.com>

# Code of Conduct

### DIFF
--- a/tests/files/das-root.json
+++ b/tests/files/das-root.json
@@ -1,70 +1,68 @@
 {
-    "root": {
-        "signatures": [
-            {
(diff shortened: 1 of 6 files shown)


tests/files/das-root.json
```diff
@@ -1,70 +1,68 @@
 {
```
MVrachev, line 1: The change in this file is small, but because the indentation was 4 and I made it to 2 it seem as my file is a totally new one.

kairoaraujo (approved):

PR #600, opened 2024-05-25
Improvements on top of new sign and tests

MVrachev:
# Description
The changes aim to:
1. fail early if none of 'api-server' or 'SIGNING_JSON_INPUT_FILE' is
provided
2. allow the combination of 'api-server' AND 'SIGNING_JSON_INPUT_FILE'
3. if 'save' is not used and 'api-server' is not provided then save
result at DEFAULT_PATH (sign_payload.json at the moment)
4. simplify the sign helper functions
5. modify the 'tests/files/payload/sign_pending_roles.json' example
payload to follow the actual API response from "GET /api/v1/metadata"

Finally, the changes address some small comments made by @lukpueh in a previous pr
and improve tests.

# Additional requirements
- [x] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

# Code of Conduct
By submitting this PR, you agree to follow our [Code of Conduct](CODE_OF_CONDUCT.rst).

- [x] I agree to follow this project's Code of Conduct

Issue #533, opened 2024-03-13
Task: review missing features of `admin2` commands


MVrachev:
### What is the task about?

Compare all missing features from `admin2` commands that exist in our current corresponding commands.

This will be an umbrella issue for missing features for each of the commands.

### Parent feature

https://github.com/repository-service-tuf/repository-service-tuf-cli/issues/532

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #602, opened 2024-05-27
Ceremony add support for API communication, tests and more

MVrachev:
**Note: This pr is rebased on top of #600.**
Merge that one before reviewing this one.

# Description

In this pr I implement some of the missing features from our new admin ceremony.

The features included here are:
- Add an `--api-server` option for communication with existing RSTUF API deployment.
- Check bootstrap status before pushing to the API.
- Push resulting payload to API `POST /api/v1/bootstrap`
- Replace `--save` option with `--output` argument.
- Automatically save if `--bootstrap` is used even if `--output` file argument is not provided.
The saved location is a default one - `ceremony_payload.json`.
- Wait for the task to finish and report back to the CLI user.
- Add `timeout` option for RSTUF API requests.
- Don't allow threshold < 2.
-  Add tests to:
1.  make sure the online key is not any of the root keys.
2. make sure at least `threshold` number of keys must be described with their public keys.

Tests for all tests are added and documentation is updated.

Signed-off-by: Martin Vrachev <martin.vrachev@broadcom.com>

Related to #533.

# Types of changes

# Additional requirements

# Code of Conduct


MVrachev:
As discussed with @kairoaraujo I have also introduced the `--offline` and `--api-server`
`admin` level options in this pr as well in my last two commits.

lukpueh:
Can we try to keep diffs to a minimum? I understand that it sometimes just isn't feasible, but this PR definitely mixes some unrelated things.

lukpueh:
Also, a consistent/accurate PR description would be helpful, e.g.:

> Automatically save if --bootstrap is used even if --output file argument is not provided.
The saved location is a default one - ceremony_payload.json.

vs.

>  This allowed the removal of "--bootstrap" in ceremony and "--api-server"
options for both "ceremony" and "sign".

lukpueh:
This is going in the right direction. Here are a few comments on args/opts:


* Moving `--api-server` to the admin base command is an improvement.
* I'm unsure about `--offline`. It is only used by 3 out of 5 admin commands, and it means different things for those:
	* For ceremony it means don't write to the API.
	* For sign and update, it means don't read from or write to the API.
* `--offline` affects the behavior of opts/args for local file input/output, in an IMO non-intuitive way. If `--offline` is passed, sign and update require an input parameter, and the output parameter no longer determines, if a file is written or not.

* Generally, the opts/args related to local files  don't feel consistent or intuitive to me.
	* `ceremony` has an optional positional `OUTPUT` path argument
	* `sign` (and likely `update`) has an optional positional `SIGNING_JSON_INPUT_FILE` path argument, and a `--save` (OUTPUT) option






**I think  we should...**

* consistently use options for both input and output files
* maybe scope `--offline` to the API write operation only, and call it `--dry-run`, then it would also feel save to not write to a local file per default.
* e.g.:

```
rstuf [--api-server] admin ceremony [--out [PATH]] [--dry-run]

Perform ceremony and send result to API to trigger bootstrap.

* If `--out [PATH]` is passed, result is written to local PATH (in addition to
  being sent to API).

* If `--dry-run` is passed, result is not sent to API. You can pass `--out
  [PATH]` to store the result locally. `--api-server` will be ignored and may
  be omitted.

```
```
rstuf [--api-server] admin {sign, metadata update} [--in PATH] [--out [PATH]] [--dry-run]

Perform sign/metadata update for pending event and send result to API.

* If `--in PATH` is passed, input is not read from API but from local PATH.

* If `--out [PATH]` is passed, result is written to local PATH (in addition to
  being sent to API).
  
* If `--dry-run` is passed, result is not sent to API. You can pass `--out
  [PATH]` to store the result locally. `--api-server` will be ignored and may
  be omitted.

```

 (I'm open to naming suggestions)

MVrachev:
#607 will replace the first 4 commits of this pr.
Please, review it before this one as this one will be on top of it.

MVrachev:
This pr is superseded by: https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/607 and  https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/612.

PR #607, opened 2024-06-04
Add tests for ceremony edge cases & improve type annotations

MVrachev:
# Description

Add:  
- test we cannot add online key as one of root keys
- requirement that threshold > 2 & add test for that
- test that expiration days must be a positive number

Additionally, improve type annotations in a few places
and move ceremony tests input inside conftest to be reused.

Related to #533

# Additional requirements

# Code of Conduct

### DIFF
--- a/repository_service_tuf/cli/admin/helpers.py
+++ b/repository_service_tuf/cli/admin/helpers.py
@@ -126,9 +126,7 @@ class _PositiveIntPrompt(IntPrompt):
 
 
 class _MoreThan1Prompt(IntPrompt):
-    validate_error_message = (
-        "[prompt.invalid]Please enter a positive number above 1"
-    )
+    validate_error_message = "[prompt.invalid]Please enter threshold above 1"
 
     def process_response(self, value: str) -> int:
         return_value: int = super().process_response(value)
--- a/repository_service_tuf/cli/admin/helpers.py
+++ b/repository_service_tuf/cli/admin/helpers.py
@@ -125,6 +125,18 @@ class _PositiveIntPrompt(IntPrompt):
         return return_value
 
 
+class _MoreThan1Prompt(IntPrompt):
+    validate_error_message = (
+        "[prompt.invalid]Please enter a positive number above 1"
+    )
+
+    def process_response(self, value: str) -> int:
+        return_value: int = super().process_response(value)
+        if return_value < 2:
+            raise InvalidResponse(self.validate_error_message)
+        return return_value
+
+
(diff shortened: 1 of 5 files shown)


MVrachev:
Not sure how this not covered line: https://app.codecov.io/gh/repository-service-tuf/repository-service-tuf-cli/pull/607/blob/repository_service_tuf/helpers/tuf.py#L272 is part of my pr, but will check if I can do something about it.

MVrachev:
I decided to just ignore it as it's part of the old md update which will be soon refactored.

kairoaraujo (approved):

MVrachev:
FT fix is required before merging this pr.
Have a look at: https://github.com/repository-service-tuf/repository-service-tuf/pull/742

lukpueh (changes requested):
This looks good. I have one inline comment, which should be addressed, and two high level comments, which aren't blockers: 

Generally, I think it's not feasible to test all code branches of the cli tool. Maybe it would be enough to just test this feature on the related helper?

If you do test the ceremony command, I recommend to store the complete list of inputs separately for each invocation, even if there is some redundancy.  Re-using some parts (e.g input_step1, input_step3) and patching others (input_step2) makes the test a bit harder to read.

MVrachev:
> This looks good. I have one inline comment, which should be addressed, and two high level comments, which aren't blockers:
> 
> Generally, I think it's not feasible to test all code branches of the cli tool. Maybe it would be enough to just test this feature on the related helper?

Well, that's a discussion worth having.
I am okay with having helper-specific tests instead of full-flow tests for those cases specific to the helpers.
For example, for threshold, we could test the helper itself only.
Still, given that the test is there and working I won't change it now.

> If you do test the ceremony command, I recommend to store the complete list of inputs separately for each invocation, even if there is some redundancy. Re-using some parts (e.g input_step1, input_step3) and patching others (input_step2) makes the test a bit harder to read.

I disagree. I prefer reading what is changed as to have all of the inputs there and wonder are there any changes in any of the steps.
When you change only for example `input2`, then you know that the change is ONLY related to the root keys configuration.

MVrachev:
Rebased on top of main.
The only new/changed commit is the last one addressing @lukpueh comment.

lukpueh (approved):

MVrachev:
Again, we need to have https://github.com/repository-service-tuf/repository-service-tuf/pull/742 merged, before ft tests pass here.

kairoaraujo (approved):

PR #612, opened 2024-06-06
Admin commands options modifications

MVrachev:
**Note: Based on top of pr #602 and umbrella pr https://github.com/repository-service-tuf/repository-service-tuf/pull/742. 
Merge those before merging this one.**

# Description

Introduce admin commands options modifications:
- add `--api-url` as a top-level `admin` command.
- add `--timeout` ceremony option
- change `ceremony OUTPUT arg` and `sign --save` into one common option `--out`.
- change `sign SIGNING_JSON_INPUT_FILE` argument to `--in`
The last two changes aim to provide options with simpler names, remove the need of multiple arguments and overall achieve consistency between `admin` commands.

With the introduction of `--api-url` as an `admin` option I have also added
RSTUF API communication for ceremony.

TODO in future prs:
- introduce `--dry-run` option (see https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/602#issuecomment-2145289938)
- save file automatically in `DEFAULT_PATH` for `ceremony` and `sign`  if `--dry-run` is set, but `--out` is not.
- after `--dry-run` is added, then add better docs for `ceremony` and `sign`

Fixes #(issue)

# Types of changes

# Additional requirements

# Code of Conduct


lukpueh:
> add --timeout ceremony option

I suggest to add this to the admin base command too. It seems relevant whenever we interact with the api.

lukpueh:
> save file automatically in DEFAULT_PATH for ceremony and sign if --dry-run is set, but --out is not.

We can discuss this when we get there, but I prefer to not save it automatically. Dry run, usually means no side-effect. Instead we can mention in the help string that `--out` should be used to persist the result. Then both options remain independent.

MVrachev:
> > add --timeout ceremony option
> 
> I suggest to add this to the admin base command too. It seems relevant whenever we interact with the API.

Good suggestion, will move it there and update the commits.

MVrachev:
I realized the way we use `--timeout` for `bootstrap` is a little strange.
Currently, we accept `--timeout` as an option for `POST /api/v1/bootstrap` https://github.com/repository-service-tuf/repository-service-tuf-api/blob/cda8e5d3366feb424c54fc4562391040f79160ae/repository_service_tuf_api/bootstrap.py#L111 and then we use it to make sure that a thread posting to the worker https://github.com/repository-service-tuf/repository-service-tuf-api/blob/cda8e5d3366feb424c54fc4562391040f79160ae/repository_service_tuf_api/bootstrap.py#L230 will finish correctly if bootstrap has succeded or if timeout has been exceeded: https://github.com/repository-service-tuf/repository-service-tuf-api/blob/cda8e5d3366feb424c54fc4562391040f79160ae/repository_service_tuf_api/bootstrap.py#L154

I think we need to discuss how we want to support `--timeout`:
- will it be `bootstrap` specific for this scenario? Maybe we can rename it to something else?
- timeout as an `requests.(get|post|delete)` option

I think both can work, but we need to rename the `bootstrap` CLI option to something else which we can then pass in the payload.

MVrachev:
Rebased on top of main and ready for review @lukpueh and @kairoaraujo.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -31,28 +31,45 @@
...
+        raise click.ClickException(
+            "Either '--api-sever'/'SERVER' in RSTUF config or '--out' needed"
+        )
 
-    if not save:  # pragma: no cover
-        _warn_no_save()
+    if settings.get("SERVER"):
+        bs_status = bootstrap_status(settings)
+        if bs_status.get("data", {}).get("bootstrap") is True:
+            raise click.ClickException(f"{bs_status.get('message')}")
```
lukpueh, line 69: Do we need to check the bootstrap status here? Wouldn't it be enough, if the server was ready for bootstrap at the time of sending the bootstrap message below? We need to handle the error case, where the server wasn't ready, anyway, because things can change between here and line 124.

IMO adding this LBYL check here is premature ux optimisation and I suggest to remove it.

lukpueh (commented):
Looks good. I left a few comments on `ceremony`, they mostly also apply to the `sign` command.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -95,11 +112,21 @@ def ceremony(save) -> None:
...
-        settings = Settings(roles)
-        payload = CeremonyPayload(settings, metadatas)
-        json.dump(asdict(payload), save, indent=2)
-        console.print(f"Saved result to '{save.name}'")
+    metadatas = Metadatas(root_md.to_dict())
+    roles_settings = Settings(roles)
+    bootstrap_payload = CeremonyPayload(roles_settings, metadatas)
+    # Dump payload when the user explicitly wants or doesn't send it to the API
+    if out:
+        save_payload(out.name, asdict(bootstrap_payload))
```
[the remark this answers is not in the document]
MVrachev: We use `save_payload` as it handles an error and it as a `click.ClickException` like this:
image

Instead of:
image

Now, whether we want to add `.json` is another question.
I think it's okay given that we check if `filename` ends with `.json`.

repository_service_tuf/cli/admin/ceremony.py
```diff
@@ -31,28 +31,45 @@
...
 )
-def ceremony(save) -> None:
+@click.pass_context
+def ceremony(context: Any, out: Optional[click.File]) -> None:
     """Bootstrap Ceremony to create initial root metadata and RSTUF config."""
     console.print("\n", Markdown("# Metadata Bootstrap Tool"))
+    settings = context.obj["settings"]
+    # Running online ceremony requires connection to the server and
+    # confirmation that the server is indeed ready for bootstap.
+    if not settings.get("SERVER") and not out:
```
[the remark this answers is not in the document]
MVrachev: I will remove it until we decide what to do with `dry-run`. This is a leftover from before as the changes were made a week or two ago.

repository_service_tuf/cli/admin/ceremony.py
MVrachev: I think it's better with the check that without it.
It's better to receive a message that bootstrap has been done already:
image

as compared to doing all of the work for the ceremony, adding at least two keys, and signing with one of them just to get:
image

lukpueh: As I said, we should handle the case (or rather the many possible cases), where the server fails, anyway. This includes giving proper user feedback and likely dumping the result to a local file so that the work is not lost.

Adding this preliminary check but not handling errors is the wrong order of priorities.

repository_service_tuf/cli/admin/ceremony.py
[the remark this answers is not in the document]
lukpueh: `save` is already a file-like object. It's a hack to take its name and open a different file object. It also bypasses all the neat [click file magic](https://click.palletsprojects.com/en/8.1.x/arguments/#file-arguments). This includes raising `click.FileError`, which would give you the desired formatting automatically.

If you want to open the file yourself, then you should make the argument a `click.Path` type instead. But I really don't see the advantage.

If you want to keep the `.json` suffix, you should change the docs.

lukpueh (changes requested):

repository_service_tuf/cli/admin/ceremony.py
MVrachev: We already handle other server problems in `bootstrap_status`:
image

repository_service_tuf/cli/admin/ceremony.py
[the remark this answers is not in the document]
MVrachev: I think for now until we don't have `--dry-run` implemented this is the correct solution as you will either use server or `--out`.
The way I see it this check will not be deleted, but will be changed - you either run online or with dry-run:
`if not settings.get("SERVER") and not dry_run` and the message will be changed as well.

repository_service_tuf/cli/admin/ceremony.py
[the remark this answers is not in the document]
MVrachev: I thought about it and given that most of the errors we could have are with file openings, then probably it makes sense to relly on click for that.
If we find other errors not handled from click we can have a helper such as `save_payload` in the future.
For now, I will not remove `save_payload` as it's used from admin-legacy.

repository_service_tuf/cli/admin/ceremony.py
lukpueh: I am talking about `send_payload`. It could fail for all sorts of reasons.  If that happens, the user's work will be lost, unless you  handle it.

MVrachev: We handle errors in `send_payload` as well:
image

lukpueh: Yes, you re-raise them and work will be lost.

MVrachev: Ahaa, I see now what you mean.
I agree saving the payload in the default path if not saved before sending to the payload sounds smart.
Will add this here as it's a small addition.

lukpueh: Not sure, if we're really on the same page. I was trying to say that the preliminary check (`bootstrap_status`) only gives you a limited certainty that posting the data (`send_payload`) will succeed. So I was actually suggesting to reduce functionality and not to add more. :)

MVrachev: I don't agree that it's an unnecessary check.
Calling `bootstrap_status` early can help the user when:
1. the given `settings.SERVER` is not reachable
2. RSTUF deployment has been bootstrapped already

For `1` he will be able to resolve his `--api-url`/`settings.SERVER` before thinking about the ceremony as a whole.
For `2` if we don't check that in the beginning and he does a full ceremony even if we save payload locally it's likely that the ceremony was a waste of time. As probably it doesn't make sense to restart a bootstrap RSTUF deployment.
Also, if `2` happens at the beginning of the ceremony then the user will be able to assess whether restarting RSTUF is required or not.

If you want to discuss it further maybe it's good to do it on a daily when it's three of us.

MVrachev: Some notes from a conversation with @lukpueh and @kairoaraujo:
- if we want to have a pre-check before running the ceremony we need to take into account other failures besides those checked now. For example, the server could become unreachable when we finish the ceremony
- one option to consider is to start a server session at the beginning of the ceremony which will be used in the end when sending the payload.

We have decided we would remove this check for now and an issue about this will be created.
Here is the issue: https://github.com/repository-service-tuf/repository-service-tuf-cli/issues/618

MVrachev:
I had to rebase as there where other prs merged, but the only new commit is the last one.
No other changes have been made.
Ready for a review @lukpueh, @kairoaraujo.

kairoaraujo (approved):

tests/conftest.py
```diff
@@ -301,22 +306,41 @@ def invoke_command(
...
+
+    api_url = None
+    if "--api-server" in args:
+        api_server_flag_index = args.index("--api-server")
+        api_url = args.pop(api_server_flag_index + 1)
+        args.remove("--api-server")
+
+    context = test_context
+    if test_context is None:
+        context = _create_test_context(api_url)
```
lukpueh, line 326: Programming tip: I'd do all of above outside of the helper, and pass only the args and context you want here.

lukpueh (approved):
LGTM.

tests/conftest.py
MVrachev: All of the above is done inside `invoke_command` as even if you don't test `--out/--save` we want to use it as we save the result of the command in `out_file.json` inside the isolated filesystem and read the result of the command from this file.. 
I want the tests to provide only args that are connected to what they are testing.

MVrachev: Something I am investigating now is how to pass `--api-server` as it's an admin option.
Meaning it's easy for me to pass options related to the command I am currently testing for example `--in` and `--out` for `sign` command, but I need to pass `--api-server` to admin - `rstuf admin --api-server <URL> send ...` which I have not found yet how to do and that's why I am settings `settings.SERVER` as a solution for now.

lukpueh: >  I am currently testing for example --in and --out for sign command,

How can you test passing `--out`, if your helper always adds it?

MVrachev: By the way, changing the context of the command seems to be the way to go as mentioned by a `click` core-developer in two issues: https://github.com/pallets/click/issues/831#issuecomment-671048863, https://github.com/pallets/click/issues/832#issuecomment-671052776.

kairoaraujo (approved):

tests/conftest.py
```diff
@@ -315,6 +315,11 @@ def invoke_command(
     else:
         out_args = ["--out", out_file_name]
 
+    # The reason why we pass '--api-server' so then we can remove it and set it
+    # as part of the context is because we cannot pass it as an 'admin' option
+    # as we are supposed to do:
+    # https://github.com/pallets/click/issues/831#issuecomment-671048863,
+    # https://github.com/pallets/click/issues/832#issuecomment-671052776.
```
lukpueh, line 322: I find it very confusing that you are inventing a new cli option (`--api-server` does exist on the subcommands)  just for calling this helper, especially when the option does exist on the base command).

tests/conftest.py
MVrachev: > > I am currently testing for example --in and --out for sign command,
> 
> How can you test passing `--out`, if your helper always adds it?

By not calling my helper. See: https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/0d73c3d7e8059bdbcb8b1722ab871c41c0c85e06/tests/unit/cli/admin/test_ceremony.py#L10

lukpueh: Sure, but we were discussing the helper.

Anyway, I have already approved the PR. So feel free to merge.

tests/conftest.py
MVrachev: I have an idea how to improve it. Will propose another commit.

lukpueh: I recommend using the `test_context` argument of your helper.

MVrachev:
I addressed the latest comment by @lukpueh.
There needs to be a change in our ft tests after that we can approve and check if ft tests will pass and if they do merge.

tests/conftest.py
```diff
@@ -315,15 +312,10 @@ def invoke_command(
     else:
         out_args = ["--out", out_file_name]
 
-    api_url = None
-    if "--api-server" in args:
-        api_server_flag_index = args.index("--api-server")
-        api_url = args.pop(api_server_flag_index + 1)
-        args.remove("--api-server")
+    if not test_context:
+        test_context = _create_test_context()
```
lukpueh, line 316: 🎉

tests/unit/cli/admin/test_import_artifacts.py
```diff
@@ -293,16 +293,18 @@ def test_import_artifacts(self):
...
             "postgresql://postgres:secret@127.0.0.1:5433",
             "--csv",
             "artifacts1of2.csv",
             "--csv",
             "artifacts2of2.csv",
         ]
-        result = invoke_command(import_artifacts.import_artifacts, [], args)
+        test_context["settings"].SERVER = "http://127.0.0.1"
+
+        result = invoke_command(
```
lukpueh, line 305: You didn't answer my question earlier. Why do you use the helper when testing import_artifacts? The helper's main purpose seems to run invoke in an isolated_filesystem, save the result to a file, and return the contents.

This does not work for import_artifacts, and you even have to explicitly disable the behavior in the helper, when called for import_artifacts.

MVrachev: I added a commit allowing the use `custom output` tests with `invoke_command`.
How does it look for you?

lukpueh (approved):
I'm okay with merging this as is and improve things as we move forward. 

Let me once more summarise the things I pointed out in previous comments which seem improvement-worthy to me:

**Consistent use of `--api-server`, `--in`, `--out` (and  `--dry-run`)**

I still think [this behaviour](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/602#issuecomment-2145289938) is the way to go eventually (note that I just fixed a mistake there)

I also think that, until we add `--dry-run`, `--api-server` should be just mandatory period.  Right now it depends on `--out` in `ceremony` (as discussed [here](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/612#discussion_r1633382852)) and on [`--in` in `sign`](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/612/files#diff-e1acd7acd514287a2830196f272d2dd79b56f018a71167ceb97e1148eed0f754R95). Besides making the future `--dry-run` implementation a bigger behaviour change than necsessary, this temporary solution also feels inconsistent.

**Testing**

The helper function design is a matter of taste. In the case of `invoke_command`, I think you are jumping through quite a few hoops  so that you can call it from everywhere (more than half of the code is case handling calling contexts).  But I also think it's not worth our time to further discuss this.

MVrachev:
I had to submit one more commit as in my last commit it seems I have not correctly changed the unit test and they were failing. I created a fix.

PR #750, opened 2024-12-09
feat: Support headers for admin `rstuf admin -H`

kairoaraujo:
# Description
Feature to support sending headers (HTTP) while using rstuf admin command.

RSTUF might be deployed behind a API Gateway or some authentication layer that requires a token or API Key or specific headers.

This feature enables users to pass headers using `--headers/-H` parameter.

This feature can be compared with `curl -H`

It supports headers such as `apikey: somekey,  Content-Type: application/json'` for example.

Examples:
```bash
rstuf admin --api-server http://api.dev.rstuf.org -H 'apikey: mykey' ceremony
```

```bash
rstuf admin --api-server http://api.dev.rstuf.org -H $TOKEN metadata sign
```

Fixes #(issue)

# Types of changes

# Additional requirements

# Code of Conduct

### DIFF
--- a/repository_service_tuf/cli/admin/__init__.py
+++ b/repository_service_tuf/cli/admin/__init__.py
@@ -13,11 +13,22 @@ from typing import Optional
 from repository_service_tuf.cli import click, rstuf
 
 
-def _set_settings(context: click.Context, api_server: Optional[str]):
+def _set_settings(
+    context: click.Context, api_server: Optional[str], headers: Optional[str]
+):
     """Set context.obj['settings'] attributes."""
     settings = context.obj["settings"]
     if api_server:
         settings.SERVER = api_server
+    if headers:
+        settings.HEADERS = dict(
+            (key.strip(), value.strip())
+            for key, value in (
+                header.split(":", 1) for header in headers.split(",")
+            )
+        )
+    else:
+        settings.HEADERS = None
 
 
 @rstuf.group()  # type: ignore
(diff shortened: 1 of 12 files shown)


kairoaraujo:
cc @trishankatdatadog

repository_service_tuf/cli/admin/__init__.py
```diff
@@ -13,11 +13,22 @@
...
+):
     """Set context.obj['settings'] attributes."""
     settings = context.obj["settings"]
     if api_server:
         settings.SERVER = api_server
+    if headers:
+        settings.HEADERS = dict(
+            (key.strip(), value.strip())
+            for key, value in (
+                header.split(":", 1) for header in headers.split(",")
```
trishankatdatadog, line 27: Should be a bit more defensive, I think. What happens if the format is even accidentally incorrect?

trishankatdatadog: I think we might want to use [HTTP Structured Fields](https://www.fastly.com/blog/improve-http-structured-headers) instead, especially with [this library](https://github.com/mnot/http-sf).

kairoaraujo: Good catch @trishankatdatadog 
However, I had a bad time using the parser.
I handle the error and catch the exception.

nit: I also like the format of `curl -H`.

repository_service_tuf/cli/admin/__init__.py
```diff
@@ -12,12 +12,32 @@
...
+def _set_settings(
+    context: click.Context, api_server: Optional[str], headers: Optional[str]
+):
     """Set context.obj['settings'] attributes."""
     settings = context.obj["settings"]
     if api_server:
         settings.SERVER = api_server
+    if headers:
+        try:
+            settings.HEADERS = dict(
```
trishankatdatadog, line 29: We can certainly improve input validation here ([issue filed](https://github.com/repository-service-tuf/repository-service-tuf-cli/issues/753)), but I think this works for now, especially if we are careful on the server-side. Thanks!

trishankatdatadog (approved):

kairoaraujo:
@MVrachev PR requires the fix in the functional tests 
https://github.com/repository-service-tuf/repository-service-tuf/pull/836

PR #431, opened 2023-10-20
Small refactoring and reuse one `get_key()` for all commands 

MVrachev:
# Description

In this pr, I did some small refactorings such as:
- coloring password inputs
- adding one `get_key()` that is reused instead of having 3 implementations
- make sure that the question asking for key name input is the same everywhere

# Additional requirements
- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

# Code of Conduct
By submitting this PR, you agree to follow our [Code of Conduct](CODE_OF_CONDUCT.rst).

- [ ] I agree to follow this project's Code of Conduct

PR #918, opened 2026-04-04
fix: use public key filename instead of keyid for online key URI

### LINKED ISSUE #642: Other: do not use keyid as online private key file name
### What do you want to share with us?

When adding a new online key in the ceremony (and metadata update) CLI, the default keyid is assigned as private key URI (file name) of the online key:

https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/a35b820bb196c33c648ccc1fa5314de0d921d53d/repository_service_tuf/cli/admin/helpers.py#L338-L339

https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/a35b820bb196c33c648ccc1fa5314de0d921d53d/repository_service_tuf/cli/admin/helpers.py#L347-L347

This assumes that the user runs the CLI, notes down the name, and then deploys the online private key file.

However, the CLI's default behavior is to ultimately send the updated metadata to the RSTUF API. This triggers online signing, which requires the online key to already be deployed. So the key file name should be known before running the CLI.

**Fix**

An easy fix would be to ask the user for the private key filename, under which the user has deployed the key.

- we could re-use the already provided public key file name (together with some basic `.pub`-suffix stripping):
  https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/a35b820bb196c33c648ccc1fa5314de0d921d53d/repository_service_tuf/cli/admin/helpers.py#L157
- we could also re-use the key name:
 https://github.com/repository-service-tuf/repository-service-tuf-cli/blob/a35b820bb196c33c648ccc1fa5314de0d921d53d/repository_service_tuf/cli/admin/helpers.py#L335-L336

kudos to @simi for pointing out the flaw in the workflow

### References

See in https://github.com/repository-service-tuf/repository-service-tuf/issues/580 (Option 2) for context about private key URIs.

Coordinate with https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/608, which adds support for HashiCorp Vault private key URIs to the CLI.

### Code of Conduct

Sandipmandal25:
## Problem

During the TUF ceremony, `_load_online_key_prompt()` was deriving the online key URI as `fn:<keyid>`. The worker's `FileNameSigner` loads the private key by **filename** using the `fn:` scheme, not by keyid.

This meant the URI written into the root metadata pointed to a file that did not exist, causing signing to fail silently on first use.

fixes #642 also ref [#431](https://github.com/repository-service-tuf/repository-service-tuf-worker/issues/431)
## Changes

**`helpers.py`**

* `_load_key_from_file_prompt()` return type changed from `SSlibKey` to `Tuple[SSlibKey, str]` — now returns both the key and the file path
* Root key caller updated to unpack with `key, _ = ...`
* Online key caller updated to unpack with `key, path = ...`, derives
  `key_name = os.path.splitext(os.path.basename(path))[0]`, and writes
  `uri = f"fn:{key_name}"`
* Guard added: raises `ValueError` if the filename stem is empty

**`test_helpers.py`**

* `test_load_key_from_file_prompt` updated to unpack tuple and assert path returned
* `test_load_online_key_prompt_success_cases` KEY_PEM case updated: mock returns `(key, "/path/to/online.pub")`, expected URI is `"fn:online"`
* `test_load_online_key_prompt_key_already_in_use` mock updated to return tuple
* New: `test_load_online_key_prompt_key_pem_empty_filename` — verifies empty filename stem raises and returns `(None, None)`

**note** 
Worker pr  [#845 ](https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/845) fixes the dict mutation `(.pop("keyid")` modifying shared state) and adds a clear error message when `x-rstuf-online-key-uri `is missing
CLI pr [#918](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/918) — fixes the ceremony writing the key identifier instead of `fn:<filename> `into `x-rstuf-online-key-uri`, so the worker can actually find the key file at signing time

### DIFF
--- a/repository_service_tuf/cli/admin/helpers.py
+++ b/repository_service_tuf/cli/admin/helpers.py
@@ -354,8 +354,8 @@ def _prompt_public_key() -> str:
     return _prompt_key(message, file_filter)
 
 
-def _load_key_from_file_prompt() -> SSlibKey:
-    """Prompt for path to public key, return Key."""
+def _load_key_from_file_prompt() -> Tuple[SSlibKey, str]:
+    """Prompt for path to public key, return Key and path."""
 
     path = _prompt_public_key()
 
@@ -365,7 +365,7 @@ def _load_key_from_file_prompt() -> SSlibKey:
     crypto = load_pem_public_key(public_pem)
     key = SSlibKey.from_crypto(crypto)
 
-    return key
+    return key, path
 
 
 def _load_key_from_hsm_prompt() -> SSlibKey:
@@ -447,7 +447,7 @@ def _load_key_prompt(
         key: SSlibKey | SigstoreKey
         match signer_type:
             case ROOT_SIGNERS.KEY_PEM:
-                key = _load_key_from_file_prompt()
+                key, _ = _load_key_from_file_prompt()
             case ROOT_SIGNERS.SIGSTORE:
                 key = _load_key_from_sigstore_prompt()
             case ROOT_SIGNERS.HSM:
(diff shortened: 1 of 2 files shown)


Sandipmandal25:
hi @kairoaraujo plese go though this for the complete context for this 2 prs https://gist.github.com/Sandipmandal25/62f82cf60385ced0a5cd4a7d99851c47

PR #930, opened 2026-04-09
Use uv in replacement of pipenv for dependency management

### LINKED ISSUE #929: Task: Use uv to replace pipenv to manage dependencies
### What is the task about?

Use uv to manage dependencies instead of pipenv.

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

OmBiradar:
Fixes #929 

Parent issue https://github.com/repository-service-tuf/repository-service-tuf/issues/951

Adding uv adds the `pyproject.toml` and `uv.lock` in replacement of `pipfile` and `pipfile.lock` which was used by pipenv.

Confirmation:

All the tests pass


OmBiradar:
Almost done

Have to either fix the merge's or just open a new PR

tox.ini
```diff
@@ -73,5 +75,5 @@ commands =
 python =
     3.10: py310,pep8,lint,typing,precommit,test,docs
     3.11: py311,pep8,lint,typing,precommit,test,docs
-    3.12: py312,pep8,lint,typing,precommit,test,docs
-    3.13: py313,pep8,lint,typing,precommit,test,docs    
\ No newline at end of file
+    3.12: py311,pep8,lint,typing,precommit,test,docs
+    3.13: py311,pep8,lint,typing,precommit,test,docs
```
MVrachev, line 79: Why did you do this change like that? 
```suggestion
    3.12: py312,pep8,lint,typing,precommit,test,docs
    3.13: py313,pep8,lint,typing,precommit,test,docs
```

tox.ini
```diff
@@ -1,22 +1,22 @@
 [tox]
 isolated_build = true
+requires = tox-uv
 envlist = py310,py311,py312,py313,lint,typing,precommit,test
-skipsdist = true
+skipsdist = false
 
 [testenv]
+extras = sqlalchemy,psycopg2
```
MVrachev, line 8: tox-uv’s default runner (uv-venv-runner) does uv pip install and ignores the lockfile. 
To install from uv.lock you have to opt in:

```suggestion
runner = uv-venv-lock-runner
extras = sqlalchemy,psycopg2
```

On main, pipenv install used Pipfile.lock. Here, pyproject.toml pins almost nothing ("click", "rich", "tuf", etc.), so CI/tox will resolve whatever is newest on PyPI each run. 
The lockfile only helps local uv sync and the Read the Docs export.

CONTRIBUTING.rst
```diff
@@ -77,27 +74,18 @@ you stay up-to-date with our repository:
...
 
-Install the requirements from the Pipfile.
 
-The flag -d will install the development requirements:
+Create a virtual environment and install all dependencies (including dev):
 
 .. code:: shell
 
-    $ pipenv install -d
+    $ uv sync --dev
```
MVrachev, line 88: uv sync installs tools into .venv. It does not put that .venv on your PATH.

So after a contributor runs only what the docs say:
```
uv sync --dev
make precommit
```

the shell still uses system Python. make precommit runs pre-commit, make reformat runs black / isort, and the tests section runs tox. Those binaries live in .venv/bin, so you get command not found unless they happen to be installed globally.

Pipenv worked differently. pipenv shell activated the project environment, so later commands used that env’s pre-commit, tox, and rstuf. The uv docs dropped that step and did not replace it.

I suggest fixing with:

```
uv sync --all-extras
source .venv/bin/activate   # Windows: .venv\Scripts\activate
make precommit
rstuf
```

MVrachev: Also,  `uv sync --dev` does not install extras

`sqlalchemy` and `psycopg2` are optional extras. 

Tox correctly adds extras = sqlalchemy,psycopg2, but a developer following CONTRIBUTING will not get them. 

tests/unit/cli/admin/test_import_artifacts.py does import sqlalchemy.

Use something like `uv sync --all-extras`. The `--dev` flag is redundant because dev is already uv’s default group.

.github/dependabot.yml
```diff
@@ -11,7 +11,7 @@ updates:
       include: "scope"
     open-pull-requests-limit: 10
     groups:
-      pipenv:
+      uv:
         patterns:
           - "*"    
   # Monitor Github Actions
```
MVrachev, line 17: It's not for this file, but:

Pre-commit still runs tox -e requirements

`.pre-commit-config.yaml`

 calls tox -e requirements / make requirements.

There is no such tox env and no Makefile target (this was already true on main).

For uv, the replacement is:
```
[testenv:requirements]
skipsdist = true
allowlist_externals = uv
commands = uv lock --check
```

MVrachev (changes requested):

tox.ini
OmBiradar: this seems to a silly mistake from my side, will remove the diff, ig the only meaning full addition here would be to remove the last tab.

OmBiradar:
The error in tests seems to be a failed build of a dependency:
```log
× Failed to build `pykcs11==1.5.18`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit status: 1)
```
Pinning a correct version of the library in the `uv.lock` file would resolve this I guess.

OmBiradar:
I did a matrix test over py 3.10 to py 3.14

Error given on:
- 3.10
- 3.11
- 3.12
- 3.13

The failing symbol, PyInt_FromLong, is part of the Python 2 C API. Python 3 removed it long ago.
pykcs11 fixed this in 1.5.19, which added SWIG 4.5 support.

OmBiradar:
All the tests pass!!! Yayy :confetti_ball:

